PR #71, opened 2023-03-10
Becatrue/66/reorganize swift

BECATRUE:
I implemented creation and destruction of `App` and `Bus` without `dataclasses`.

### DIFF
--- a/swift/swift.py
+++ b/swift/swift.py
@@ -110,9 +110,9 @@ class Swift(QObject):
         # create an app
         cls = getattr(module, info["class"])
         if "args" in info:
-            app = cls(name, **info["args"], parent=self)
+            app = cls(name, parent=self, **info["args"])
         else:
-            app = cls(name, self)
+            app = cls(name, parent=self)
         # set a slot of broadcast signal to router
         app.broadcastRequested.connect(self._routeToBus)
         # add the app to the list of subscribers on each bus
--- a/swift/swift.py
+++ b/swift/swift.py
@@ -57,6 +57,12 @@ class Swift(QObject):
         self._show_frame(setup_env["app"])
 
     def createBus(self, name: str):
+        """Creates a global bus using set-up environment.
+        
+        Args:
+            name: A string that indicates the name of the bus.
+              It should be defined in the set-up file.
+        """
         info = self.setup_bus[name]
         # create a bus
         if "timeout" in info:
@@ -69,6 +75,37 @@ class Swift(QObject):
         # store the bus
         self._buses[name] = bus
         self._subscribers[name] = []
+
+    def createApp(self, name: str, show: str = True, pos: str = "", args: dict = None):
+        info = self.setup_app[name]
+        # import the app module
+        path = info.get("path", ".")
+        mod_name, cls_name = info["module"], info["class"]
+        with _add_to_path(os.path.dirname(path)):
+            module = importlib.import_module(mod_name)
+        # create an app
+        cls = getattr(module, cls_name)
+        args = info.get("args", {})
+        app = cls(name, self, **args)
+        # set a slot of broadcast signal to router
+        app.broadcastRequested.connect(self._routeToBus)
+        # add the app to the list of subscribers on each bus
+        for bus_name in info["bus"]:
+            self._subscribers[bus_name].append(app)
+        # show frames if the "show" option is true
+        if info["show"]:
+            for frame in app.frames():
+                dockWidget = QDockWidget(name, self.mainWindow)
+                dockWidget.setWidget(frame)
+                area = {
+                    "left": Qt.LeftDockWidgetArea,
+                    "right": Qt.RightDockWidgetArea,
+                    "top": Qt.TopDockWidgetArea,
+                    "bottom": Qt.BottomDockWidgetArea
+                }.get(info["pos"], Qt.AllDockWidgetAreas)
+                self.mainWindow.addDockWidget(area, dockWidget)
+        # store the app
+        self._apps[name] = app
 
     def _init_bus(self, setup_bus: dict):
         """Initializes global buses using set-up environment.
@@ -150,7 +187,7 @@ class Swift(QObject):
         bus.write(msg)
 
     @pyqtSlot(str)
-    def _route_to_app(self, msg: str):
+    def _routeToApp(self, msg: str):
         """Routes a signal from a bus to the apps that subscribe to it.
 
         This is a slot for the received signal of each bus.
--- a/swift/swift.py
+++ b/swift/swift.py
@@ -110,7 +110,7 @@ class Swift(QObject):
         # create an app
         cls = getattr(module, info["class"])
         if "args" in info:
-            app = cls(name, parent=self, **info["args"])
+            app = cls(name, **info["args"], parent=self)
         else:
             app = cls(name, self)
         # set a slot of broadcast signal to router
@@ -139,8 +139,9 @@ class Swift(QObject):
         Args:
             name: A name of the bus to destroy.
         """
-        self._buses[name].deleteLater()
-        self._buses.pop(name).stop()
+        bus = self._buses.pop(name)
+        bus.stop()
+        bus.deleteLater()
 
     def destroyApp(self, name: str):
         """Destroys an app.
@@ -148,10 +149,10 @@ class Swift(QObject):
         Args:
             name: A name of the app to destroy.
         """
-        self._apps[name].deleteLater()
         app = self._apps.pop(name)
         for apps in self._subscribers.values():
             apps.discard(app)
+        app.deleteLater()
 
     @pyqtSlot(str, str)
     def _routeToBus(self, busName: str, msg: str):
--- a/swift/swift.py
+++ b/swift/swift.py
@@ -31,17 +31,6 @@ class Swift(QObject):
         1. Load setup environment.
         2. Create buses.
         3. Create apps and show their frames.
-
-    Attributes:
-        _buses: A dictionary for storing buses.
-          Each element represents a bus.
-          A key is a bus name and its value is a Bus object.
-        _apps: A dictionary for storing apps.
-          Each element represents an app.
-          A key is an app name and its value is an App object.
-        _subscribers: A dictionary for storing subscribers.
-          Each element represents subscribers of each bus.
-          A key is a bus name and its value is a set containing its subscribers.
     """
 
     def __init__(self, setupEnv: dict, parent=None):
@@ -106,29 +95,29 @@ class Swift(QObject):
               path: A path desired to be added for importing app.
               module: A name of app module.
               class: A name of app class.
-              show: Whether its frames is shown at the beginning.
+              show: Whether its frames are shown at the beginning.
                 True if you want to show the frames, otherwise False.
               pos: A string that indicates the position of the frames.
                 It should be one of "left", "right", "top", or "bottom"; and is case-sensitive.
                 Otherwise, it will be regarded as default (AllDockWidgetAreas).
-              bus: A list of buses which app subscribes to.
-              args: Additional arguments for custom.
+              bus: A list of buses which the app subscribes to.
+              args: Additional arguments for app class constructor.
                 If there is no additional arguments, set None.
         """
         # import the app module
-        with _add_to_path(os.path.dirname(info.get("path", ""))):
+        with _add_to_path(os.path.dirname(info.get("path", "."))):
             module = importlib.import_module(info["module"])
         # create an app
         cls = getattr(module, info["class"])
         if "args" in info:
-            app = cls(name, self, **info["args"])
+            app = cls(name, parent=self, **info["args"])
         else:
             app = cls(name, self)
         # set a slot of broadcast signal to router
         app.broadcastRequested.connect(self._routeToBus)
         # add the app to the list of subscribers on each bus
-        for bus_name in info["bus"]:
-            self._subscribers[bus_name].add(app)
+        for busName in info["bus"]:
+            self._subscribers[busName].add(app)
         # show frames if the "show" option is true
         if info["show"]:
             for frame in app.frames():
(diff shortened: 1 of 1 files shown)


swift/swift.py
```diff
@@ -28,22 +28,31 @@ class Swift(QObject):
...
+    Attributes:
+        _buses: A dictionary for storing buses.
+          Each element represents a bus.
+          A key is a bus name and its value is a Bus object.
+        _apps: A dictionary for storing apps.
+          Each element represents an app.
+          A key is an app name and its value is an App object.
+        _subscribers: A dictionary for storing subscribers.
+          Each element represents subscribers of each bus.
+          A key is a bus name and its value is a set containing its subscribers.
```
kangz12345, line 44: We don't need to make docstrings for the internal attributes (whose names starting with _).
If you need to clarify some significant assumptions to those internal attributes, please use inline comments.
The docstring is for public API.

swift/swift.py
```diff
@@ -28,22 +28,31 @@ class Swift(QObject):
...
-              In "app", there may be keys below; 
-                "path", "module", "class", "show", "pos", "bus", and "args".
-              In "bus", there may be keys below;
-                "timeout".
-
+              In "app", there are apps containing keys as below; 
+                "path" (optional), "module", "class", "show", "pos", "bus", and "args" (optional).
+              In "bus", there are buses containing keys as below;
+                "timeout" (optional).
               For details, see setup.json.
```
kangz12345, line 56: I think this does not specify the dictionary structure enough.
For example, what are "apps" here:

> In "app", there are apps containing keys as below;

One may deduce that "apps" are `dict`s, but how are they stored in "app" section? Are they in a `list`?
Please modify the description so that a user can make a setup dictionary without the setup.json file.
Note that the setup.json file is NOT a part of the `swift` package.

kangz12345: However, maybe this is not in the scope of this PR.
I will make a new issue from this comment, and resolve it in another PR (probably in the PR regarding to the `Info`s).

swift/swift.py
```diff
@@ -52,102 +61,131 @@ def __init__(self, setup_env: dict, parent=None):
...
 
-    def _init_bus(self, setup_bus: dict):
-        """Initializes global buses using set-up environment.
+    def load(self, appInfos: dict, busInfos: dict):
+        """Initializes swift system and loads the apps and buses.
         
         Args:
-            setup_bus: Set-up environment about bus.
+            appInfos: See "app" part of setupEnv in self.__init__().
+            busInfos: See "bus" part of setupEnv in self.__init__().
```
kangz12345, line 72: 👍

swift/swift.py
```diff
@@ -52,102 +61,131 @@ def __init__(self, setup_env: dict, parent=None):
...
+        # create a bus
+        if "timeout" in info:
+            bus = Bus(name, info["timeout"])
+        else:
+            bus = Bus(name)
+        # set a slot of received signal to router
+        bus.received.connect(self._routeToApp)
+        bus.start()
+        # store the bus
+        self._buses[name] = bus
```
kangz12345, line 97: I think it should be checked first that there is already a bus with name `name`.
If there were a bus already, it will be garbage collected in python, but not in Qt, and the queue consumer thread will not stop.
Maybe returning `False` when failed would be helpful.

swift/swift.py
```diff
@@ -52,102 +61,131 @@ def __init__(self, setup_env: dict, parent=None):
...
+        self._subscribers.setdefault(name, set())
```
kangz12345, line 98: I think [`setdefault()`](https://docs.python.org/3/library/stdtypes.html#dict.setdefault) does not fit here that well.
It is used when one wants to get the value corresponding to a key, i.e. `name` here, and if the key is not in it, then add the key with the default value.
So it is similar to `get()` with default, but `get()` does not add the default value into the dict.

In this case, there must not exist `name` as I said in another comment. Therefore, using `update()` or `self._subscribers[name] = set()` would be better.

kangz12345: BTW it is a good choice to use `set` 👍

swift/swift.py
```diff
@@ -52,102 +61,131 @@ def __init__(self, setup_env: dict, parent=None):
...
+        """Creates an app and shows their frames using set-up environment.
         
         Args:
-            setup_app: Set-up environment about app.
+            name: A name of app.
+            info: A dictionary containing app info. Each element is like below:
+              path: A path desired to be added for importing app.
+              module: A name of app module.
+              class: A name of app class.
+              show: Whether its frames is shown at the beginning.
```
kangz12345, line 109: Grammar: is -> are

swift/swift.py
```diff
@@ -52,102 +61,131 @@ def __init__(self, setup_env: dict, parent=None):
...
+                True if you want to show the frames, otherwise False.
+              pos: A string that indicates the position of the frames.
+                It should be one of "left", "right", "top", or "bottom"; and is case-sensitive.
+                Otherwise, it will be regarded as default (AllDockWidgetAreas).
+              bus: A list of buses which app subscribes to.
```
kangz12345, line 114: Grammar: which app -> which the app

swift/swift.py
```diff
@@ -52,102 +61,131 @@ def __init__(self, setup_env: dict, parent=None):
...
-                    area = pos_to_area.get(info["pos"], Qt.AllDockWidgetAreas)
-                    self.mainWindow.addDockWidget(area, dockWidget)
-        self.mainWindow.show()
+        # import the app module
+        with _add_to_path(os.path.dirname(info.get("path", ""))):
+            module = importlib.import_module(info["module"])
+        # create an app
+        cls = getattr(module, info["class"])
+        if "args" in info:
+            app = cls(name, self, **info["args"])
```
kangz12345, line 124: Maybe `self` should be given as `parent=self` since we don't know the order..?

swift/swift.py
```diff
@@ -52,102 +61,131 @@ def __init__(self, setup_env: dict, parent=None):
...
+        else:
+            app = cls(name, self)
+        # set a slot of broadcast signal to router
+        app.broadcastRequested.connect(self._routeToBus)
+        # add the app to the list of subscribers on each bus
+        for bus_name in info["bus"]:
+            self._subscribers[bus_name].add(app)
```
kangz12345, line 131: camel case for `busName` ;)

swift/swift.py
kangz12345: Or, the `name` existence checking should be done before calling this method... It seems a design choice.

swift/swift.py
```diff
@@ -52,102 +61,131 @@ def __init__(self, setup_env: dict, parent=None):
...
+        # store the app
+        self._apps[name] = app
+
+    def destroyBus(self, name: str):
+        """Destroys a global bus.
+        
+        Args:
+            name: A name of the bus to destroy.
+        """
+        self._buses[name].deleteLater()
```
kangz12345, line 153: It should be deleted from Python (`self._buses.remove(name)` or something, so it can be garbage-collected).
Moreover, the bus should be stopped. Therefore, `self._buses.pop(name).stop()` would be perfect.
Finally, the `_subscribers` should be updated as well (remove the key `name`).

swift/swift.py
```diff
@@ -52,102 +61,131 @@ def __init__(self, setup_env: dict, parent=None):
...
+
+    def destroyApp(self, name: str):
+        """Destroys an app.
+        
+        Args:
+            name: A name of the app to destroy.
+        """
+        self._apps[name].deleteLater()
```
kangz12345, line 161: Here as well, the app instance should be removed from `self._apps` and the removed app should be removed from the `_subscribers` as well (from the buses which the app subscribe to).

kangz12345 (changes requested):

swift/swift.py
BECATRUE: Ok! I remove it.

swift/swift.py
```diff
@@ -52,102 +61,131 @@ def __init__(self, setup_env: dict, parent=None):
...
-            # show frames if the 'show' option is true
-            if info["show"]:
-                for frame in self._apps[name].frames():
-                    dockWidget = QDockWidget(name, self.mainWindow)
-                    dockWidget.setWidget(frame)
-                    area = pos_to_area.get(info["pos"], Qt.AllDockWidgetAreas)
-                    self.mainWindow.addDockWidget(area, dockWidget)
-        self.mainWindow.show()
+        # import the app module
+        with _add_to_path(os.path.dirname(info.get("path", ""))):
```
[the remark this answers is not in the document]
BECATRUE: Wow, that was a detailed review! It was just a mistake..

swift/swift.py
BECATRUE: No! It is defined in `app.BaseApp`.

```python
__init__(self, name: str, parent=None)
```

But, if you think it is ambiguous, I will change it into `parent=self`.

swift/swift.py
BECATRUE: Okay! I will leave it as it is.

swift/swift.py
BECATRUE: I see.. Actually, I'm going to talk about this.

In `Swift`, We only check key existence in a few cases.
Then, I suggest a solution; at the beginning part of `load()`, check if `appInfo` and `busInfo` have a correct structure.

swift/swift.py
kangz12345: I meant that `cls` which inherits `BaseApp` may take other arguments before `parent`, e.g.,
```python
class Foo(BaseApp):
  def __init__(self, name, arg1, arg2, parent=None):
    super().__init__(name, parent)
    ...
```

swift/swift.py
BECATRUE: Umm.. How about checking `name` existence and if exists, just print a message?

swift/swift.py
BECATRUE: Then, how do you think about the below situation?

1. Using `load()`, initialize `Swift`.
2. Using `destroyBus()`, destroy a bus.
3. Using `createBus()`, create a bus which I removed step 2.

According to your suggestion, the subscribers info would be removed..

kangz12345: Oh, I thought the subscriptions expire when a bus is destroyed.

swift/swift.py
kangz12345: I think it should notify the caller the failure, with an exception or return value.
Printing a message is good (you can do this along with notifying the caller), but it only notifies the user, not the method caller.

swift/swift.py
BECATRUE: Ah, then what about your idea? I think the subscriptions must not expire..

swift/swift.py
BECATRUE: It's hard for me to understand your suggestion..😂
Did you mean that printing a message is enough?

swift/swift.py
BECATRUE: Ah, I thought we restricted the order of arguments.

I will change it into `parent=self`.

swift/swift.py
BECATRUE: Ok, I will! But we should decide how to deal with `_subscribers`.

BECATRUE:
I applied reviews as much as I can now.

swift/swift.py
kangz12345: Since `parent` is an optional argument (which has a default value), we cannot force the subclasses to keep the order of (`name`, `parent=None`, `arg1`, `arg2`, ...).

swift/swift.py
kangz12345: Hmm, okay then for now, let's keep your decision, since there is no way to re-subscribe the buses at the moment.
This should be discussed further in another discussion.

swift/swift.py
kangz12345: 😄 sorry for unclarity, my opinion is:
Printing a message is okay, but in addition to that, raising an exception or returning the result is necessary.

swift/swift.py
```diff
@@ -106,29 +95,29 @@ def createApp(self, name: str, info: dict):
...
-        with _add_to_path(os.path.dirname(info.get("path", ""))):
+        with _add_to_path(os.path.dirname(info.get("path", "."))):
             module = importlib.import_module(info["module"])
         # create an app
         cls = getattr(module, info["class"])
         if "args" in info:
-            app = cls(name, self, **info["args"])
+            app = cls(name, parent=self, **info["args"])
         else:
             app = cls(name, self)
```
kangz12345, line 115: Now I realized that it should be `parent=self` here as well.
Suppose the constructor of `cls` looks like
```python
def __init__(self, name, arg1=None, arg2=None, parent=None):
```
and `"args"` is not given in order to use the default values for `arg1` and `arg2`.

swift/swift.py
```diff
@@ -151,6 +140,7 @@ def destroyBus(self, name: str):
             name: A name of the bus to destroy.
         """
         self._buses[name].deleteLater()
+        self._buses.pop(name).stop()
```
kangz12345, line 143: Hmm, maybe changing the order could be safer:
```python
self._buses[name].stop()
self._buses.pop(name).deleteLater()
```

kangz12345 (changes requested):

swift/swift.py
BECATRUE: I made some modification to implement `destroyBus()` and `destroyApp()` similarly.

BECATRUE:
I applied all reviews!

swift/swift.py
kangz12345: I was talking about the line 115.
The line 113 is not a problem since the keyword argument order is not important.
However, as in line 115, if the argument is given as a positional argument, the `self` in the second position could be misgiven as another argument (`arg1`, not `parent` in my example above).

kangz12345 (changes requested):
Just one thing left..

BECATRUE:
> Just one thing left..

Oh, I understand now!

BECATRUE:
I updated it! Please check.

@kangz12345

kangz12345 (approved):

PR #80, opened 2023-03-13
Improved handling of app and bus information introducing dataclasses `AppInfo` and `BusInfo`

### LINKED ISSUE #72: Detailed setup dictionary structure documentation
Setup dictionary structure documentation should be clear enough so that users can build it without the "setup.json" file, since it is not a part of `swift` package.
It should be removed soon (maybe should be moved into examples folder).

See also: _https://github.com/snu-quiqcl/swift/pull/71#discussion_r1132430009_

### LINKED ISSUE #73: [FEATURE] AppInfo and BusInfo for structuring the information of Apps and Buses
Add `AppInfo` and `BusInfo` classes in `swift.swift` module to provide a helpful API to deal with the essential information of Apps and Buses.
This will support more confident creation of Apps and Buses, even for the other applications in the future.

Use [`dataclass`](https://docs.python.org/3/library/dataclasses.html#dataclasses.dataclass) to easily express what attributes are required to create an App or a Bus.

A brief picture would be:
```python
@dataclass
class AppInfo:
  """...Docstring for attributes..."""
  module: str
  class_: str # class is a reserved keyword
  path: str | None = None
  show: bool = True
  pos: str = ""
  bus: Iterable[str] = ()
  args: Mapping[str, Any] | None = None

@dataclass
class BusInfo:
  """...Docstring for attributes..."""
  timeout: float | None = None

def info_str(info: AppInfo | BusInfo) -> str:
  """Converts an Info into a JSON string."""
  ...
```

From a JSON string `s`, an `AppInfo` object can be created by:
```python
info = AppInfo(**json.loads(s))
```
### Additional context
This will be developed after #66 and #71.
This will resolve #72.

kangz12345:
[`dataclasses.dataclass`](https://docs.python.org/3/library/dataclasses.html#dataclasses.dataclass) is useful when dealing with structured data with default values, etc.
I applied this to apps and buses initialization information in `swift`.
This makes things easier to check if the data has a proper structure.
Moreover, I implemented converter functions (`parse()` and `strinfo()`).

In addition, I removed some comments that seems merely repeating the codes.

Please freely give me your opinions!

Currently, pylint complains about type hints.
The main reason is that [`typing.Self`](https://docs.python.org/3/library/typing.html#typing.Self) is introduced in Python 3.11.
Also, the `|` in type hints is introduced in Python 3.10.
I will look for the workaround for this issue...

This closes #72 and closes #73.

### DIFF
--- a/swift/swift.py
+++ b/swift/swift.py
@@ -16,11 +16,40 @@ import json
 import importlib
 import importlib.util
 from contextlib import contextmanager
+from dataclasses import dataclass
+from typing import Any, Iterable, Mapping
 
 from PyQt5.QtCore import QObject, pyqtSlot, Qt
 from PyQt5.QtWidgets import QApplication, QMainWindow, QDockWidget
 
 from swift.bus import Bus
+
+@dataclass
+class AppInfo:
+    """Information required to create an app.
+    
+    Fields:
+        module: Module name in which the app class resides.
+        class_: App class name.
+        path: System path for module importing. None if it is not necessary.
+        show: Whether to show the app frames on creation.
+        pos: Position on the main window; refer to Qt.DockWidgetArea enum.
+          Should be one of "left", "right", "top", or "bottom", case-sensitive.
+          Otherwise, defaults to Qt.AllDockWidgetAreas.
+        bus: Buses which the app subscribes to.
+        args: Keyword argument dictionary of the app class constructor.
+          It must exclude name and parent arguments. Even if they exist, they will be ignored.
+          None for initializing the app with default values,
+          where only name and parent arguments will be passed.
+    """
+    module: str
+    class_: str
+    path: str | None = None
+    show: bool = True
+    pos: str = ""
+    bus: Iterable[str] = ()
+    args: Mapping[str, Any] | None = None
+
 
 class Swift(QObject):
     """Actual manager for swift system.
--- a/swift/swift.py
+++ b/swift/swift.py
@@ -17,7 +17,7 @@ import importlib
 import importlib.util
 from contextlib import contextmanager
 from dataclasses import dataclass, asdict
-from typing import Any, Iterable, Mapping
+from typing import Any, Iterable, Mapping, Self
 
 from PyQt5.QtCore import QObject, pyqtSlot, Qt
 from PyQt5.QtWidgets import QApplication, QMainWindow, QDockWidget
@@ -49,6 +49,30 @@ class AppInfo:
     pos: str = ""
     bus: Iterable[str] = ()
     args: Mapping[str, Any] | None = None
+
+    @classmethod
+    def parse(cls, info: str) -> Self:
+        """Constructs an AppInfo object from a JSON string.
+
+        This does not catch JSONDecodeErrors since it clearly requires the JSON string
+        for the input.
+        Care must be taken not to input a non-JSON formatted string.
+        
+        Args:
+            info: A JSON string of a dictionary that contains the information of an app.
+              Its keys are field names of AppInfo and values are corresponding values.
+              Exceptionally for "class_" field, "class" is also accepted.
+              If both "class_" and "class" exist, then "class" is ignored.
+              If both does not exist, a KeyError("class_") is raised.
+        
+        Raises:
+            KeyError: When there is no mandatory fields in info.
+        """
+        info: dict[str, Any] = json.loads(info)
+        class_ = info.pop("class", None)
+        if info.setdefault("class_", class_) is None:
+            raise KeyError("class_")
+        return cls(**info)
 
 
 @dataclass
--- a/swift/swift.py
+++ b/swift/swift.py
@@ -83,7 +83,7 @@ class Swift(QObject):
         3. Create apps and show their frames.
     """
 
-    def __init__(self, setupEnv: dict, parent: QObject | None = None):
+    def __init__(self, setupEnv: Mapping[str, Mapping], parent: QObject | None = None):
         """
         Args:
             setupEnv: A dictionary containing set-up environment about app and bus.
@@ -103,7 +103,7 @@ class Swift(QObject):
         self.load(setupEnv["app"], setupEnv["bus"])
         self.mainWindow.show()
 
-    def load(self, appInfos: dict, busInfos: dict):
+    def load(self, appInfos: Mapping[str, Mapping], busInfos: Mapping[str, Mapping]):
         """Initializes swift system and loads the apps and buses.
         
         Args:
@@ -115,7 +115,7 @@ class Swift(QObject):
         for name, info in appInfos.items():
             self.createApp(name, info)
 
-    def createBus(self, name: str, info: dict):
+    def createBus(self, name: str, info: Mapping[str, Any]):
         """Creates a global bus using set-up environment.
         
         Args:
@@ -136,7 +136,7 @@ class Swift(QObject):
         self._buses[name] = bus
         self._subscribers.setdefault(name, set())
 
-    def createApp(self, name: str, info: dict):
+    def createApp(self, name: str, info: Mapping[str, Any]):
         """Creates an app and shows their frames using set-up environment.
         
         Args:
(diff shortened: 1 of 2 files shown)


kangz12345:
If you think this PR is too large, I will split this into two.

BECATRUE:
First of all, please check Pylint warning!

swift/swift.py
```diff
@@ -16,12 +16,95 @@
...
 
 from swift.bus import Bus
 
+@dataclass
+class AppInfo:
+    """Information required to create an app.
+    
+    Fields:
+        module: Module name in which the app class resides.
+        class_: App class name.
```
BECATRUE, line 33: I think `cls` looks better according to PEP.

In the [documentation](https://peps.python.org/pep-0008/#designing-for-inheritance),
> If your public attribute name collides with a reserved keyword, append a single trailing underscore to your attribute name. This is preferable to an abbreviation or corrupted spelling. (However, notwithstanding this rule, ‘cls’ is the preferred spelling for any variable or argument which is known to be a class, especially the first argument to a class method.)

How about your opinion?

swift/swift.py
```diff
@@ -33,93 +116,77 @@ class Swift(QObject):
...
-    def load(self, appInfos: dict, busInfos: dict):
+    def load(self, appInfos: Mapping[str, AppInfo], busInfos: Mapping[str, BusInfo]):
         """Initializes swift system and loads the apps and buses.
         
         Args:
-            appInfos: See "app" part of setupEnv in self.__init__().
-            busInfos: See "bus" part of setupEnv in self.__init__().
+            appInfos: A dictionary whose keys are app names and the values are
+              corresponding AppInfo objects. All the apps in the dictionary
+              will be created and shown if the show field is True.
```
BECATRUE, line 146: It looks a little ambiguous since it may be interpretted that 'will be created if the show field is True'.

Is there any better explanation..?

swift/swift.py
```diff
@@ -16,12 +16,95 @@
...
+        Args:
+            info: A JSON string of a dictionary that contains the information of an app.
+              Its keys are field names of AppInfo and values are corresponding values.
+              Exceptionally for "class_" field, "class" is also accepted.
+              If both "class_" and "class" exist, then "class" is ignored.
+              If both does not exist, a KeyError("class_") is raised.
+        
+        Raises:
+            KeyError: When there is no mandatory fields in info.
+        """
```
BECATRUE, line 66: Like other functions with returning value, how about adding 'Return' explanation?

swift/swift.py
```diff
@@ -221,30 +287,50 @@ def _get_argparser():
...
+    See AppInfo and its parse() for app_info_* structure.
+    See BusInfo and its parse() for bus_info_* structure.
+      
     Args:
         setup_path: A path of set-up file.
 
     Returns:
-        dict: A dictionary containing set-up environment about app and bus.
-          For details, see Swift.__init__().
+        (app_infos, bus_infos): Dictionaries of set-up information about app and bus.
```
BECATRUE, line 313: Umm, isn't it a tuple?

BECATRUE (commented):
It looks so good!
I left a few comments.

kangz12345:
> First of all, what do you think of updating Python version to 3.11?

Hmm, it's really tempting to do so, but at the same time, I am worried about potential bugs, since Python 3.11 is quite new.

[Status of Python Versions](https://devguide.python.org/versions/) would be helpful reference to decide this.

[Python 3.10](https://peps.python.org/pep-0619/) seems stable enough (in terms of bugs) since the bugfix release schedule is almost ended.

I am not familiar to such version stability issues, so we have to search for good references about this.

Please create an issue about this and let's discuss it further.

swift/swift.py
```diff
@@ -16,12 +16,95 @@
...
 from PyQt5.QtCore import QObject, pyqtSlot, Qt
 from PyQt5.QtWidgets import QApplication, QMainWindow, QDockWidget
 
 from swift.bus import Bus
 
+@dataclass
+class AppInfo:
+    """Information required to create an app.
+    
+    Fields:
```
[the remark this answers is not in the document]
kangz12345: It was hard for me to choose the word 😄
I decided to use "Fields" because `dataclasses` use the word [`field`](https://docs.python.org/3/library/dataclasses.html#dataclasses.field) for the data class attributes.

In addition, they are not "Args" but "Attributes".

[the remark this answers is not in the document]
kangz12345: This decision is valid exclusively for data classes.
If you have other opinions, freely suggest other options.

swift/swift.py
kangz12345: What a great review! I couldn't find that, thank you 👍

swift/swift.py
kangz12345: Oh, I meant (a tuple of) dictionaries. Does it look ambiguous or misleading? If so, I will change the expression.

swift/swift.py
kangz12345: How do you think of accepting "class" as "cls" in `AppInfo.parse()`?
It is convenient but could be confused.

swift/swift.py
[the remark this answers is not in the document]
BECATRUE: If you think `Fields` is more appropriate, I'm ok!

swift/swift.py
BECATRUE: Oh, I agree. There is already `cls` variable..

swift/swift.py
BECATRUE: I can understand what you intend, but for newcomers, changing the expression looks better.

swift/swift.py
kangz12345: I think the summary tells us enough information about the returning value.
Or should I change the summary to start with "Returns an App Info ..." instead of "Constructs an AppInfo ..."?

BECATRUE:
> > First of all, what do you think of updating Python version to 3.11?
> 
> Hmm, it's really tempting to do so, but at the same time, I am worried about potential bugs, since Python 3.11 is quite new.
> 
> [Status of Python Versions](https://devguide.python.org/versions/) would be helpful reference to decide this.
> 
> [Python 3.10](https://peps.python.org/pep-0619/) seems stable enough (in terms of bugs) since the bugfix release schedule is almost ended.
> 
> I am not familiar to such version stability issues, so we have to search for good references about this.
> 
> Please create an issue about this and let's discuss it further.

I agree with you very much. Let's create a new issue.

swift/swift.py
kangz12345: Oh, I changed "class_" -> "cls". I think you misunderstood my question.

I just want your opinion about the current decision of `AppInfo.parse()`.
It takes "class" as well as "cls" (previously "class_"), since the previous setup.json file uses "class".
My question is: should it allow "class" or not (i.e., strictly allow "cls" only)?

swift/swift.py
kangz12345: Alright, I changed the expression!

kangz12345:
I resolved the Pylint warnings regarding to type hints.
See #89 for more details.

swift/swift.py
BECATRUE: Umm.. Do you want to set a variable name as "class"?
Is is possible even if "class" is a reseved word?

swift/swift.py
BECATRUE: I think that writing like "Returns an App Info ..." is better to understand!

swift/swift.py
kangz12345: No, the field name is "cls".

However, we give a dictionary as a setup information.
Its keys may contain "class".
Should this be accepted as "cls"?

BECATRUE: To understand easily, I think we should only allow `cls`.

Thus, please change it and update `setup.json` too.

kangz12345:
Please check @BECATRUE @Ahrrri

BECATRUE (approved):
Wow. Python Typing looks so cool!

I approved the PR.

PR #87, opened 2023-03-16
Update README.md and write Wiki

BECATRUE:
Move all descriptions to wiki and write how to use swift.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,9 +12,6 @@ One of the below:
 1. `git clone ${url}`
 2. In the repository, `pip install -e .`
 
-For convenience, you may download test databases:
-- [database.zip](https://github.com/snu-quiqcl/swift/files/10907031/database.zip)
-
 ## How to use
 In the repository, just do like as below:
 
--- a/README.md
+++ b/README.md
@@ -3,10 +3,14 @@
 # swift
 Swift (**S**NU **w**idget **i**ntegration **f**ramework for PyQ**t**) is a framework for integration of PyQt widgets where they can communicate with each other. This project is mainly developed for trapped ion experiment controller GUI in SNU QuIQCL.
 
-## How to download
+## How to download and install
 One of the below:
-- git clone or fetch
-- Download the latest release: [v1.0.1](https://github.com/snu-quiqcl/swift/releases/tag/v1.0.1)
+- Using released version (_recommended_)
+1. `pip install git_https://github.com/snu-quiqcl/swift.git@v1.0.1`
+
+- Using git directly
+1. `git clone ${url}`
+2. In the repository, `pip install -e .`
 
 For convenience, you may download test databases:
 - [database.zip](https://github.com/snu-quiqcl/swift/files/10907031/database.zip)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,149 +1,17 @@
...
 # swift
-**S**NU **w**idget **i**ntegration **f**ramework for PyQ**t**
+Swift (**S**NU **w**idget **i**ntegration **f**ramework for PyQ**t**) is a framework for integration of PyQt widgets where they can communicate with each other. This project is mainly developed for trapped ion experiment controller GUI in SNU QuIQCL.
 
-A framework for integration of PyQt widgets where they can communicate with each other.
-This project is mainly developed for trapped ion experiment controller GUI in SNU QuIQCL.
+## How to download
+One of the below:
+- git clone or fetch
+- Download the latest release: [v1.0.1](https://github.com/snu-quiqcl/swift/releases/tag/v1.0.1)
```
kangz12345, line 9: I think this should be separated into two: download and install.
Could you please make an additional instruction of installing the package using `pip`?

README.md
```diff
@@ -1,149 +1,17 @@
...
 
-`swift` provides a dashboard-like main window, in which variety of `Frame`s reside.
-A `Frame` is in fact a special `QWidget` which obeys the interface of `swift`, and it will be wrapped by a `QDockWidget`.
+For convenience, you may download test databases:
+- [database.zip](https://github.com/snu-quiqcl/swift/files/10907031/database.zip)
```
kangz12345, line 12: Please add more comments on what this is.
Actually, since the example is not a part of the package in fact, I think this should be in example section somewhere in the Wiki.
What do you think of this? Should it be here in README.md?

README.md
```diff
@@ -1,149 +1,17 @@
...
-Not required.
-
-## 5. Data calculator
-### GUI
-- A combobox for selecting the database from which the value of 'A' is fetched
-- A combobox for selecting the database from which the value of 'B' is fetched
-- A label showing the sum of recently fetched 'A' and 'B', or an error message if something goes wrong
-### Backend
-- `read(db) -> int`: Read the given database and return the fetched number
+`python -m swift.swift`.
```
kangz12345, line 17: Please add the use case with `-s` option.

kangz12345 (changes requested):

README.md
BECATRUE: Oh, I completely forgot it😂

BECATRUE: I think it is little hard to separate into two, because using released version, we can download and install, simultaneously. So, I just separated in two part: using released version and using git directly.

README.md
BECATRUE: I agree with you.

Then, how about moving to the discussion? Actually, it's already there.

BECATRUE:
I updated it!

README.md
kangz12345: > Then, how about moving to the discussion? Actually, it's already there.

Great, then let's remove it from README.md.

README.md
kangz12345: Good decision!

kangz12345 (changes requested):
Please just remove the example database part.

BECATRUE:
I removed it!

kangz12345 (approved):

PR #91, opened 2023-03-16
Set a central widget in the mainWindow

### LINKED ISSUE #85: Inserting a central Widget within a main window
In the current toy model, we have some uncomfortable problems with `QDockWidget`s.

For example, if we take all the `QDockWidget` out of the `QMainWindow` and try to place them into the main window again, 
we can't place them in all directions: only top/bottom or left/right.

At first, I found another option using `QMdiArea` and `QMdiSubWindow`s, but this still has some bothering kinds of stuff.

Finally, there seems to be a solution while retaining our `QDockWidget`s layout: setting a central widget with `setCentralWidget()`.
If we set a `QWidget` as a central widget, then it seems to work well as I expect.

So the main issue is: Which widget should be placed in the center?

This issue continues from #79

Ahrrri:
I just set a central widget with a `QLabel` named "swift".
This closes #85

### DIFF
--- a/swift/swift.py
+++ b/swift/swift.py
@@ -48,6 +48,8 @@ class Swift(QObject):
         super().__init__(parent=parent)
         self.mainWindow = QMainWindow()
         self.centralWidget = QLabel("Swift")
+        self.centralWidget.setAlignment(Qt.AlignCenter)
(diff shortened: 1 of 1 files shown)


kangz12345 (approved):
LGTM

PR #94, opened 2023-03-17
Fix the error that the bus thread is not quited 

BECATRUE:
So far, when the thread is finished, we only quit it.
However, the application is stopped so quickly, there is no enough time for bus thread to finish quit.
And it cause the error: `Destroyed while thread is still running`.

Thus, I just called `wait()` after `quit()`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,9 +12,6 @@ One of the below:
 1. `git clone ${url}`
 2. In the repository, `pip install -e .`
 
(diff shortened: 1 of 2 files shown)


kangz12345:
Please explain what `wait()` does or provide a link of documentation.

In addition, I don't want this PR changes README.md.
Is this branch created from the new version of `main` branch?

PR #103, opened 2023-03-20
Type hints

### LINKED ISSUE #78: Using type hints
### Discussed in https://github.com/snu-quiqcl/swift/discussions/75

<sup>Originally posted by **kangz12345** March 11, 2023</sup>
As you can see in [PEP 484 - Type Hints](https://peps.python.org/pep-0484), type hints can improve static analysis of the codes.
Moreover, I think the code would be clearer and less documentation will be required.

How about forcing to use type hints in our repo?</div>

### LINKED ISSUE #89: Old version compatible type annotation
Since the type annotation syntax is quite new in Python, we have troubles using the latest type annotation syntax.
Since we use Python 3.8-3.10 for Pylint checking, we should support at least Python 3.8.
Moreover, since our main computer in the lab is using Python 3.7, it seems better to support Python 3.7 as well.

1. [Forward references](https://peps.python.org/pep-0484/#forward-references)
If we use the type that is not defined yet, e.g., in the class definition itself, we cannot use the class name as a type since it is not defined. In that case, we use a string literal as you see in the linked reference.

2. OR (`|`) operator on type annotations
We can use [`Union`](https://docs.python.org/3.8/library/typing.html#typing.Union) for this.
In addition, [`Optional`](https://docs.python.org/3.8/library/typing.html#typing.Optional) seems useful for `Union` with `None`.

3. `list`, `dict`, `tuple`, etc.
In the old versions, we can use `List`, `Dict`, `Tuple`, etc. instead.

This will be applied in #80.
See also #78 and #84.

kangz12345:
I annotated all the type hints of functions and methods.
The static type checkers, e.g. Pylance in VSCode?, will now check your code more thoroughly.

As a side effect, in a good way, a potential bug in `datacalc` is resolved.
See also: _[commit 379fa80](https://github.com/snu-quiqcl/swift/pull/103/commits/379fa8038835f86aababe0e55aec1dace49f9000)_.

This closes #78.
This resolves #89.

### DIFF
--- a/apps/backend.py
+++ b/apps/backend.py
@@ -5,6 +5,7 @@ Backend module for offering various functions.
 import random
 import time
 import sqlite3
+from typing import Any
 
 def generate() -> int:
     """Generates a random number from 0 to 99.
@@ -24,7 +25,7 @@ def poll() -> int:
     return int(time.time()) % 100
 
 
-def read(db_path: str, table: str):
+def read(db_path: str, table: str) -> Any:
     """Reads the value from the database.
 
     It can only read the last row in a specific table.
--- a/apps/backend.py
+++ b/apps/backend.py
@@ -63,7 +63,7 @@ def read(db_path: str, table: str) -> Any:
     return value
 
 
-def write(db_path: str, table: str, value) -> bool:
+def write(db_path: str, table: str, value: Any) -> bool:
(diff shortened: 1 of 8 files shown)


BECATRUE (approved):
It looks perfect!

PR #104, opened 2023-03-21
Update README.md

BECATRUE:
I realized that the README file in `develop` branch is not updated to be the same as `main` branch.
Thus, I modified it.


kangz12345:
How about rebasing `develop` onto `main`?
Which way would be safer?

BECATRUE:
> How about rebasing `develop` onto `main`?
> Which way would be safer?

Oh, I think it is safer.

I will update this branch later.

PR #108, opened 2023-03-21
Add Python 3.7 runner for Pylint

### LINKED ISSUE #84: Python version choice
> First of all, what do you think of updating Python version to 3.11?

Hmm, it's really tempting to do so, but at the same time, I am worried about potential bugs, since Python 3.11 is quite new.

[Status of Python Versions](https://devguide.python.org/versions/) would be helpful reference to decide this.

[Python 3.10](https://peps.python.org/pep-0619/) seems stable enough (in terms of bugs) since the bugfix release schedule is almost ended.

I am not familiar to such version stability issues, so we have to search for good references about this.

_Originally posted by @kangz12345 in https://github.com/snu-quiqcl/swift/issues/80#issuecomment-1467353805_

kangz12345:
Since the computers in the lab use Python 3.7, we need to make the codes compatible with that version.

This closes #84.

### DIFF
--- a/.github/workflows/pylint.yml
+++ b/.github/workflows/pylint.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10"]
+        python-version: ["3.7", "3.8", "3.9", "3.10"]
     steps:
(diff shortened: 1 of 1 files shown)


BECATRUE (approved):
LGTM

PR #109, opened 2023-03-21
`parent` parameter order to the last

### LINKED ISSUE #98: `parent` parameter for `QObject`s
Currently, there are many `QObject`s we have implemented, and each of them has the `parent` attribute since it is the one of `QObject`.

However, some of them do not have `parent` parameter at all (e.g., `Bus`), and the position of `parent` in the constructor is not consistent.

Therefore, I want to suggest that we put `parent` parameter at the last order, i.e., `def __init__(self, ..., parent=None)`.

1. Add `parent` parameter if there is not.
2. Change the position of `parent` parameter to the last.

See also: #30

kangz12345:
I moved the `parent` parameter to the last parameter of constructors.
Moreover, `Bus` and `QueueConsumer` didn't have the `parent`, so I added it.
In addition, I changed to only using keyword argument passing for `parent`, i.e., `__init__(arg1, ..., parent=parent)`.

This closes #98.

### DIFF
--- a/examples/datacalc.py
+++ b/examples/datacalc.py
@@ -62,7 +62,7 @@ class DataCalcApp(BaseApp):
         Args:
             tables: See DataCalcApp.tables.
         """
-        super().__init__(name, parent)
+        super().__init__(name, parent=parent)
         self.tables = tables
         self.dbs = {"": ""}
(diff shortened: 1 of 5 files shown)


BECATRUE (approved):
LGTM

PR #116, opened 2023-03-24
Remove statements about sys.modules in _add_to_path()

### LINKED ISSUE #113: Why do we even consider 'sys.modules' in 'swift.swift._add_to_path'?
### Discussed in https://github.com/snu-quiqcl/swift/discussions/111

<sup>Originally posted by **BECATRUE** March 23, 2023</sup>
I have a question about `_add_to_path()` function in `swift.swift` module.

The function is as below:
```python
@contextmanager
def _add_to_path(path: str):
    """Adds a path temporarily.

    Using a 'with' statement, you can import a module without changing sys.path.

    Args:
        path: A desired path to be added. 
    """
    old_modules = sys.modules
    sys.modules = old_modules.copy()
    old_path = sys.path
    sys.path = old_path.copy()
    sys.path.insert(0, path)
    try:
        yield
    finally:
        sys.path = old_path
        sys.modules = old_modules
```

In the function, we change only `sys.path`, but why do we even care about `sys.modules`?

I think it is not neccessary to consider `sys.modules`.</div>

I just want to remove the part about `sys.modules`.

BECATRUE:
This PR will close #113.

I just removed three statements about `sys.modules`.

### DIFF
--- a/swift/swift.py
+++ b/swift/swift.py
@@ -251,8 +251,6 @@ def _add_to_path(path: str):
     Args:
         path: A desired path to be added. 
     """
-    old_modules = sys.modules
-    sys.modules = old_modules.copy()
     old_path = sys.path
     sys.path = old_path.copy()
     sys.path.insert(0, path)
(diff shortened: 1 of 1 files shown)


kangz12345 (approved):
LGTM

PR #122, opened 2023-03-27
Test swift.swift

### LINKED ISSUE #107: [FEATURE] Unit test for swift module
I want to implement the unit test for `swift.swift` module.

#### Test for `parse()`
- TBD

#### Test for `strinfo()`
- TBD

#### Test for `Swift` class
- TBD

#### Test for `_add_to_path()`
- If it doesn't change `sys`.

#### Test for `_get_argparser()`
- If the parser can parse command line arguments.

#### Test for `_read_setup_file()`
- If a setup file is opened well.
- If `app_infos` and `bus_infos` are assigned using the setup file.

#### Test for `main()`
- If each statements is run well.

### Additional context
It is necessary to use `mock` module.

See also: [mock](https://docs.python.org/ko/3/library/unittest.mock.html#module-unittest.mock)

BECATRUE:
This PR will close #107.

I implemented all tests in `swift.swift`.

During tests, I found a minor bug in `Swift` class.
In `createApp()`, we added `QDockWidget`s to `self.mainWindow` as below:
```python
area = {
    "left": Qt.LeftDockWidgetArea,
    "right": Qt.RightDockWidgetArea,
    "top": Qt.TopDockWidgetArea,
    "bottom": Qt.BottomDockWidgetArea
}.get(info.pos, Qt.AllDockWidgetAreas)
self.mainWindow.addDockWidget(area, dockWidget)
```
However, if `area` arguments is `Qt.AllDockWidgetAreas`, the warning occurs like below:
```console
QMainWindow::addDockWidget: invalid 'area' argument
```
Thus, as a solution, I changed it into `Qt.LeftDockWidgetArea`.

If you think it should be separated into new issue, I will do so.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -9,3 +9,6 @@ variable-rgx=_?_?(?:(?:[a-z0-9]+(?:_[a-z0-9]+)*_?_?)|(?:[a-z0-9]+(?:[A-Z][a-z0-9
 
 [SIMILARITIES]
 min-similarity-lines=20
+
+[MESSAGES CONTROL]
+disable=protected-access
--- a/tests/test_swift.py
+++ b/tests/test_swift.py
@@ -8,7 +8,6 @@ import sys
 import os
 import importlib
 from collections.abc import Iterable
-import json
 
 from PyQt5.QtCore import QObject
 from PyQt5.QtWidgets import QApplication
--- a/tests/test_swift.py
+++ b/tests/test_swift.py
@@ -5,7 +5,6 @@ Module for testing swift package.
 import unittest
 from unittest.mock import MagicMock, patch
 import sys
-import os
 import importlib
 from collections.abc import Iterable
 
@@ -145,7 +144,6 @@ class SwiftFunctionTest(unittest.TestCase):
         """Test _get_argparser()."""
         parser = swift._get_argparser()
         args = parser.parse_args()
-        self.assertEqual(parser.description, "SNU widget integration framework for PyQt")
         self.assertEqual(args.setup_path, "test_setup.json")
 
     @patch.object(sys, "argv", [""])
@@ -163,7 +161,7 @@ class SwiftFunctionTest(unittest.TestCase):
         mock_load.assert_called_once()
 
     @patch("swift.swift._get_argparser")
-    @patch("swift.swift._read_setup_file", return_value=({}, {}))
+    @patch("swift.swift._read_setup_file", return_value={})
     @patch("swift.swift.Swift")
     @patch("PyQt5.QtWidgets.QApplication.exec_")
     def test_main(self, mock_get_argparser, mock_read_setup_file, mock_swift, mock_exec_):
--- a/tests/test_swift.py
+++ b/tests/test_swift.py
@@ -10,7 +10,7 @@ import importlib
 from collections.abc import Iterable
 
 from PyQt5.QtCore import QObject
-from PyQt5.QtWidgets import QApplication
+from PyQt5.QtWidgets import QApplication, QMainWindow, QLabel
 
 from swift.app import BaseApp
 from swift.swift import (
@@ -92,6 +92,14 @@ class SwiftTest(unittest.TestCase):
 
     def test_init(self):
         """Test if swift is initialized correctly."""
+        self.assertIsInstance(self.swift.mainWindow, QMainWindow)
+        self.assertIsInstance(self.swift.centralWidget, QLabel)
+        buses = set()
+        for name, appInfo in APP_INFOS.items():
+            self.assertIn(name, self.swift._apps)
+            buses.update(appInfo.bus)
+        for bus in buses:
+            self.assertIn(bus, self.swift._subscribers)
 
 
 class SwiftFunctionTest(unittest.TestCase):
--- a/tests/test_swift.py
+++ b/tests/test_swift.py
@@ -14,6 +14,52 @@ from PyQt5.QtWidgets import QApplication
 from swift.app import BaseApp
 from swift.swift import Swift, AppInfo, BusInfo, _read_setup_file
 
+APP_INFOS = {
+    "app1": AppInfo(
+        module="module1",
+        cls="cls1",
+        path="path1",
+        show=False,
+        pos="left",
+        bus=["bus1", "bus2"],
+        args={"arg1": "value1"}
+    ),
+    "app2": AppInfo(
+        module="module2",
+        cls="cls2"
+    )
+}
+
+APP_DICTS = {
+    "app1": {
+        "module": "module1",
+        "cls": "cls1",
+        "path": "path1",
+        "show": False,
+        "pos": "left",
+        "bus": ["bus1", "bus2"],
+        "args": {"arg1": "value1"}
+    },
+    "app2": {
+        "module": "module2",
+        "cls": "cls2"
+    }
+}
+
+BUS_INFOS = {
+    "bus1": BusInfo(
+        timeout=5.0
+    ),
+    "bus2": BusInfo()
+}
+
+BUS_DICTS = {
+    "bus1": {
+        "timeout": 5.0
+    },
+    "bus2": {}
+}
+
 class AppTest(unittest.TestCase):
     """Unit test for app.py."""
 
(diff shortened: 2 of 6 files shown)


BECATRUE:
The testing result is like below:

image

It is also shown in GitHub Action details.

tests/test_swift.py
```diff
@@ -29,5 +78,100 @@ def test_frames(self):
...
+        name = list(APP_INFOS.keys())[0]
+        self.swift.destroyApp(name)
+        self.assertNotIn(name, self.swift._apps)
+
+    def test_broadcast(self):
+        """Test _broadcast()."""
+        busName = list(self.buses)[0]
+        self.swift._broadcast(busName, "test_msg")
+        for app in self.swift._subscribers[busName]:
+            app.received.emit.assert_called_once()
```
kangz12345, line 118: This seems not clear, because the testing process is not quite predictable or straightforward.
1. `set` iterator would not follow an intuitive order (which bus name will be tested?)
2. What if there is no subscribers for the selected bus?

Therefore, please make it more explicit that which bus name will be tested and how many subscribers there are.

tests/test_swift.py
```diff
@@ -29,5 +78,100 @@ def test_frames(self):
...
+
+
+class SwiftFunctionTest(unittest.TestCase):
+    """Unit test for functions in swift.py"""
+
+    def test_parse(self):
+        """Test parse()."""
+        self.assertEqual(parse(AppInfo, APP_JSONS["app1"]), APP_INFOS["app1"])
+        self.assertEqual(parse(AppInfo, APP_JSONS["app2_default"]), APP_INFOS["app2"])
```
kangz12345, line 127: How about breaking this into two unit tests, i.e., `test_parse()` and `test_parse_default()`?

tests/test_swift.py
```diff
@@ -29,5 +78,100 @@ def test_frames(self):
...
+    def test_strinfo(self):
+        """Test strinfo()."""
+        self.assertEqual(strinfo(APP_INFOS["app1"]), APP_JSONS["app1"])
+        self.assertEqual(strinfo(APP_INFOS["app2"]), APP_JSONS["app2"])
+
+    def test_add_to_path(self):
+        """Test _add_to_path()."""
+        old_path = sys.path.copy()
+        with _add_to_path(os.path.dirname("test_dir")):
+            self.assertNotEqual(old_path, sys.path)
```
kangz12345, line 138: Maybe checking if they are not equal is not enough to test this function.
In addition to that, please check if the test directory is _in_ `sys.path`.

tests/test_swift.py
```diff
@@ -29,5 +78,100 @@ def test_frames(self):
...
+        self.assertEqual(old_path, sys.path)
+
+    @patch.object(sys, "argv", ["", "-s", "test_setup.json"])
+    def test_get_argparser(self):
+        """Test _get_argparser()."""
+        parser = _get_argparser()
+        args = parser.parse_args()
+        self.assertEqual(parser.description, "SNU widget integration framework for PyQt")
```
kangz12345, line 146: I think this is not necessary, since the description can be changed and it does nothing with API.

tests/test_swift.py
```diff
@@ -29,5 +78,100 @@ def test_frames(self):
...
+    @patch("builtins.open")
+    @patch("json.load", return_value={"app": APP_DICTS})
+    def test_read_setup_file(self, mock_open, mock_load):
+        """Test _read_setup_file()."""
+        self.assertEqual(_read_setup_file(""), APP_INFOS)
+        mock_open.assert_called_once()
+        mock_load.assert_called_once()
+
+    @patch("swift.swift._get_argparser")
+    @patch("swift.swift._read_setup_file", return_value=({}, {}))
```
kangz12345, line 164: `_read_setup_files()` now returns only one dict. Why is this working..?

tests/test_swift.py
```diff
@@ -29,5 +78,100 @@ def test_frames(self):
...
+    @patch("swift.swift.Swift")
+    @patch("PyQt5.QtWidgets.QApplication.exec_")
+    def test_main(self, mock_get_argparser, mock_read_setup_file, mock_swift, mock_exec_):
+        """Test main()."""
```
kangz12345, line 168: Hmm.. the docstrings seems too trivial.
Should we omit the docstring for testing unless there is additional context should be noted?
Please give me your opinion.

kangz12345 (changes requested):

kangz12345:
In addition, why isn't there a test for `createApp()`?

.pylintrc
```diff
@@ -9,3 +9,6 @@ variable-rgx=_?_?(?:(?:[a-z0-9]+(?:_[a-z0-9]+)*_?_?)|(?:[a-z0-9]+(?:[A-Z][a-z0-9
 
 [SIMILARITIES]
 min-similarity-lines=20
+
+[MESSAGES CONTROL]
+disable=protected-access
```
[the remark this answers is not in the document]
BECATRUE: Oh, it looks better.

Then, I will change like below:
1. Remove `protected-access` in the pylintrc file.
2. Make another pylintrc file in `tests` dir with `protected-access` option.
3. In Pylint Action, process two pylint commands: one is for files except to `tests` dir and the other is for `tests`.

tests/test_swift.py
```diff
@@ -3,11 +3,60 @@
...
 from collections.abc import Iterable
 
 from PyQt5.QtCore import QObject
+from PyQt5.QtWidgets import QApplication, QMainWindow, QWidget, QLabel
 
 from swift.app import BaseApp
+from swift.swift import (
+    Swift, AppInfo, parse, strinfo,
+    _add_to_path, _get_argparser, _read_setup_file, main
+)
```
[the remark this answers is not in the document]
BECATRUE: Thanks for reference! I think importing the module is more clear.

tests/test_swift.py
```diff
@@ -29,5 +78,100 @@ def test_frames(self):
...
+        self.swift = Swift(APP_INFOS)
+
+    def test_init(self):
+        """Test if swift is initialized correctly."""
+        self.assertIsInstance(self.swift.mainWindow, QMainWindow)
+        self.assertIsInstance(self.swift.centralWidget, QLabel)
+        for name in APP_INFOS:
+            self.assertIn(name, self.swift._apps)
+        for bus in self.buses:
+            self.assertIn(bus, self.swift._subscribers)
```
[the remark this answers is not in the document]
BECATRUE: Umm, however, I think the purpose of testing is checking if our system works correctly and all variables should be tested even if it is protected.

What do you think of it?

.pylintrc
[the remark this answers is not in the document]
kangz12345: Okay, looks good!

tests/test_swift.py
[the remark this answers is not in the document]
kangz12345: Hmm yes you are right, but for the roles of internal variables, _correct_ behavior is not _specified_.
Therefore, for me, it does not make sense to test the internal variables.
Nevertheless, if we assume something, for example an invariant condition, for some internal variables, we could test that assumption.
For internal methods or functions, I understand it because we internally specify how it should work.

Rather, I think we should define getters to iterate over the apps and subscribers.

tests/test_swift.py
kangz12345: > I sorted `self.buses` with alphabetical order.

Is `set` iteration order deterministic? Note that `self.buses` is a `set`, which is an unordered collection.

> I defined `self.buses` as an union set of bus list of `AppInfo`s. So I think there is no case when the bus has no subscribers.

I meant what if `self.swift._subscribers[busName]` is empty?
Then no assertion would be performed, right?
I am worried about this because the code looks like it does not care the tested case (i.e., `busName` and its subscribers here), but it actually does (as you said, you are assuming some properties about the test case).
This may lead to mistakenly hidden bugs.
Consider, later we change the test case, e.g., `APP_INFOS`, and somehow we changed something about `_broadcast()`.
There might be a chance that there is actually a bug, but this test code does not find it, because we changed the test case breaking the assumptions of this test.

tests/test_swift.py
BECATRUE: Wow, it seems to have worked lucky. I updated it.

BECATRUE:
> In addition, why isn't there a test for `createApp()`?

I doesn't test it because it is called automatically by the constructor.
And also, I tested with two diffrent app infos, so all functions in `createApp()` were checked.

tests/test_swift.py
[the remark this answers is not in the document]
BECATRUE: Okay! Then, I will remove all testing parts about protected variabls.

BECATRUE:
First of all, I updated it up to where I modified it.

tests/test_swift.py
kangz12345: Thank you for the suggestions.
I would prefer the second option, with the pylintrc for testing.

tests/test_swift.py
BECATRUE: > Is set iteration order deterministic? Note that self.buses is a set, which is an unordered collection.

I updated `self.buses` as follows.
```python
self.buses = set()
for appInfo in APP_INFOS.values():
    self.buses.update(appInfo.bus)
self.buses = sorted(self.buses)
```
Then, it is changed into a sorted list.

BECATRUE:
I applied all reviews.

In `test_broadcast()`, I called `_broadcast()` with all buses and checked if each app is called by the number of buses it subcribes to.
In `test_destroy_app()`, I destroyed all apps and checked if `self.swift._apps` is empty.

tests/test_swift.py
```diff
@@ -101,37 +95,31 @@ def test_init(self):
             self.assertIn(bus, self.swift._subscribers)
 
     def test_destroy_app(self):
-        """Test destroyApp()."""
-        name = next(iter(APP_INFOS))
-        self.swift.destroyApp(name)
-        self.assertNotIn(name, self.swift._apps)
+        for name in APP_INFOS:
+            self.swift.destroyApp(name)
```
kangz12345, line 99: It would be better to check `NotIn` after destroying each app!

tests/test_swift.py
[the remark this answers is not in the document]
kangz12345: For now, let's stick to this implementation.
Later I will add some public methods and modify the test codes along.

kangz12345 (changes requested):
Final review!

BECATRUE:
I updated it! @kangz12345

kangz12345 (approved):
LGTM

Issue #82, opened 2023-03-14
[FEATURE] Update a part adding/removing DBs


BECATRUE:
![image](https://user-images.githubusercontent.com/76851886/224932060-067274e9-18a9-494d-96ab-2b9ee7751e34.png)

In the above logging messages, pay attention to messages of which timestamp is 16:50:49.

There are many duplicated messages, such as "Polled database is not selected.".
I just removed a dataset 'A.sqlite' in dbmgr, but it occured duplicated messages.

This occurs because each update calls the function below.
> self.viewerFrame.dbBox.clear()
> self.viewerFrame.dbBox.addItem("")

Other than this, there are several problems such that we always replace all database list to received list, and it is so inefficient.

Therefore, I want to implement as below:
- Remove duplicated messages.
- Not replacing all database list, only adding or removing changed databases.

_Originally posted by @BECATRUE in https://github.com/snu-quiqcl/swift/issues/81#issuecomment-1467311437_

kangz12345:
Please provide more information about the bug. What is the problem and how are you going to fix it?

BECATRUE:
> Please provide more information about the bug. What is the problem and how are you going to fix it?

I will write more infos later.

BECATRUE:
I updated `updateDB()` in `poller`.

Please check it! @kangz12345

kangz12345:
> I updated `updateDB()` in `poller`.
> 
> Please check it! @kangz12345

Do you mean in the commit `eda90e2` in branch `BECATRUE/82/update_db`?

BECATRUE:
> > I updated `updateDB()` in `poller`.
> > Please check it! @kangz12345
> 
> Do you mean in the commit `eda90e2` in branch `BECATRUE/82/update_db`?

Yes, it is right!

PR #124, opened 2023-03-29
Fix a duplicated called bug in pollor

BECATRUE:
This PR is related to #82, not close.

Before fix all bugs, first of all I modified `updateDB()` in `pollor`.

### Situation
1. Add a database in `dbmgr`.
2. Set the database as a store of `poller`.
3. Remove the database from `dbmgr`.

### Original
image

We can check that the polling message is duplicated.

### Fixed
image

The duplicated message is removed.

### DIFF
--- a/examples/poller.py
+++ b/examples/poller.py
@@ -87,7 +87,9 @@ class PollerApp(BaseApp):
         """Updates the database list using the transferred message.
 
         This is a slot for received signal.
-        It assumes that the new database is added at the end.
+        It assumes that:
+            The new database is always added at the end.
+            Changing the order of the databases is not allowed.
 
         Args:
             busName: A name of the bus that transfered the signal.
--- a/examples/poller.py
+++ b/examples/poller.py
@@ -87,6 +87,7 @@ class PollerApp(BaseApp):
         """Updates the database list using the transferred message.
 
         This is a slot for received signal.
+        It assumes that the new database is added at the end.
 
         Args:
             busName: A name of the bus that transfered the signal.
--- a/examples/poller.py
+++ b/examples/poller.py
@@ -67,6 +67,7 @@ class PollerApp(BaseApp):
         self.dbs = {"": ""}
         self.dbName = ""
         self.viewerFrame = ViewerFrame()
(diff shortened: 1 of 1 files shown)


BECATRUE:
After your review, a few codes were modified, so I just made a PR.

examples/poller.py
```diff
@@ -98,20 +99,24 @@ def updateDB(self, busName: str, msg: str):
...
+                newDBs = set([""])
                 for db in msg.get("db", ()):
                     if all(key in db for key in ("name", "path")):
                         name, path = db["name"], db["path"]
-                        self.dbs[name] = path
-                        self.viewerFrame.dbBox.addItem(name)
+                        newDBs.add(name)
+                        if name not in self.dbs:
+                            self.dbs[name] = path
+                            self.viewerFrame.dbBox.addItem(name)
```
kangz12345, line 110: This assumes that the new item is always added at the end, which is actually true for now, but it would be better to specify that assumption somewhere, e.g., in the docstring of this method.

kangz12345 (changes requested):

kangz12345:
Thank you for the screenshots!

examples/poller.py
BECATRUE: Okay! As you said, I assumed that the new database is added at the end and it is so weird.
In addition, it is inefficient to send the whole list from `dbmgr` whenever the database list updates.

However, it is just a toy project, I still remain it.

BECATRUE:
I updated the docstring. Please check it.

examples/poller.py
```diff
@@ -86,6 +87,7 @@ def updateDB(self, busName: str, msg: str):
         """Updates the database list using the transferred message.
 
         This is a slot for received signal.
+        It assumes that the new database is added at the end.
```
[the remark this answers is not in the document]
BECATRUE: Oh, that's no problem!

I will apply this.

BECATRUE:
I applied your reviews. @kangz12345

kangz12345 (approved):
LGTM

PR #126, opened 2023-03-31
Fix all duplicated called bugs

### LINKED ISSUE #82: [FEATURE] Update a part adding/removing DBs

BECATRUE:
This PR will close #82.

This is a follow-up to #124. Similar to the issue, I modified all `updateDB()`s in `numgen` and `datacalc`.

### Situation
1. Add database `A` and `B` in `dbmgr`.
2. Set `A` as a store of `numgen`.
3. Set `A` and `B` as sources of `datacalc`.
4. Remove `A` from `dbmgr`.

### Before step 4
image

### After step 4
image

There is no duplicated message.

### DIFF
--- a/examples/datacalc.py
+++ b/examples/datacalc.py
@@ -112,7 +112,10 @@ class DataCalcApp(BaseApp):
                     newDBs.add(name)
                     if name not in self.dbs:
                         self.dbs[name] = path
-                        list(map(operator.methodcaller("addItem", name), self.viewerFrame.dbBoxes.values()))
+                        list(map(
+                            operator.methodcaller("addItem", name),
+                            self.viewerFrame.dbBoxes.values()
+                        ))
                 removingDBs = originalDBs - newDBs
                 for dbBox in self.viewerFrame.dbBoxes.values():
                     if dbBox.currentText() in removingDBs:
--- a/examples/datacalc.py
+++ b/examples/datacalc.py
@@ -127,7 +127,11 @@ class DataCalcApp(BaseApp):
 
     @pyqtSlot(str)
     def setDB(self, name: str):
-        """Sets the database to fetch the numbers."""
+        """Sets the database to fetch the numbers.
+        
+        Args:
+            name: A name of the selected combobox.
+        """
         dbBox = self.viewerFrame.dbBoxes[name]
         self.dbNames[name] = dbBox.currentText()
         self.broadcastRequested.emit(
--- a/examples/datacalc.py
+++ b/examples/datacalc.py
@@ -5,7 +5,6 @@ App module for showing the sum of two values from selected databases.
 import os
 import json
 import functools
-import operator
 from typing import Optional, Dict, Tuple
 
 from PyQt5.QtCore import QObject, pyqtSlot
@@ -112,10 +111,8 @@ class DataCalcApp(BaseApp):
                     newDBs.add(name)
                     if name not in self.dbs:
                         self.dbs[name] = path
-                        list(map(
-                            operator.methodcaller("addItem", name),
-                            self.viewerFrame.dbBoxes.values()
-                        ))
+                        for dbBox in self.viewerFrame.dbBoxes.values():
+                            dbBox.addItem(name)
                 removingDBs = originalDBs - newDBs
                 for dbBox in self.viewerFrame.dbBoxes.values():
                     if dbBox.currentText() in removingDBs:
--- a/examples/datacalc.py
+++ b/examples/datacalc.py
@@ -84,6 +84,9 @@ class DataCalcApp(BaseApp):
         """Updates the database list using the transferred message.
 
         This is a slot for received signal.
+        It assumes that:
+            The new database is always added at the end.
+            Changing the order of the databases is not allowed.
 
         Args:
             channelName: A name of the channel that transfered the signal.
--- a/examples/datacalc.py
+++ b/examples/datacalc.py
@@ -67,10 +67,12 @@ class DataCalcApp(BaseApp):
         self.dbs = {"": ""}
         self.dbNames = {"A": "", "B": ""}
         self.viewerFrame = ViewerFrame()
+        for dbBox in self.viewerFrame.dbBoxes.values():
+            dbBox.addItem("")
         # connect signals to slots
         self.received.connect(self.updateDB)
-        for dbBox in self.viewerFrame.dbBoxes.values():
-            dbBox.currentIndexChanged.connect(self.setDB)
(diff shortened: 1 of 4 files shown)


BECATRUE:
I will re-open this PR after resolving Pylint warnings.

examples/datacalc.py
```diff
@@ -67,10 +67,12 @@ def __init__(self, name: str, tables: Dict[str, str], parent: Optional[QObject]
...
         self.dbNames = {"A": "", "B": ""}
         self.viewerFrame = ViewerFrame()
+        for dbBox in self.viewerFrame.dbBoxes.values():
+            dbBox.addItem("")
         # connect signals to slots
         self.received.connect(self.updateDB)
-        for dbBox in self.viewerFrame.dbBoxes.values():
-            dbBox.currentIndexChanged.connect(self.setDB)
+        self.viewerFrame.dbBoxes["A"].currentIndexChanged.connect(lambda: self.setDB("A"))
+        self.viewerFrame.dbBoxes["B"].currentIndexChanged.connect(lambda: self.setDB("B"))
```
BECATRUE, line 75: I set a slot of each combobox individually, because the following code doesn't work correctly.
```python
for dbName, dbBox in self.viewerFrame.dbBoxes.items():
    dbBox.currentIndexChanged.connect(lambda: self.setDB(dbName))
```

In the code, two slots are set as `lambda: self.setDB("B")`, not separately.
I think it may be a problem in terms of reference and I tried to fix it, but I couldn't make it.

kangz12345: In that case, you can use [`functools.partial()`](https://docs.python.org/3/library/functools.html#functools.partial):
```python
functools.partial(self.setDB, dbName) # instead of lambda
```
If you use the `lambda` expression like that, it actually defines a function, and hence refers to `dbName` _when the function is called_.
Therefore the both dbBox A and B will be connected to technically the same slots, since the both slots refer to the same variable `dbName`.

However, if you use `functools.partial()`, the _value_ of `dbName` is passed when `partial()` function is called (i.e., just before the slot is connected).

examples/datacalc.py
```diff
@@ -94,38 +99,42 @@ def updateDB(self, busName: str, msg: str):
...
                     else:
                         print(f"The message was ignored because "
                               f"the database {db} has no such key; name or path.")
-                for name, orgDbName in orgDbNames.items():
-                    if orgDbName in self.dbs:
-                        self.viewerFrame.dbBoxes[name].setCurrentText(orgDbName)
+                removingDBs = originalDBs - newDBs
+                for dbBox in self.viewerFrame.dbBoxes.values():
+                    if dbBox.currentText() in removingDBs:
+                        dbBox.setCurrentText("")
```
kangz12345, line 118: Why do you need to make the text ""?

examples/datacalc.py
```diff
@@ -94,38 +99,42 @@ def updateDB(self, busName: str, msg: str):
...
-            self.dbNames[name] = dbBox.currentText()
-            self.broadcastRequested.emit(
-                "logbus", 
-                f"Database {name} is set as {self.dbNames[name]}."
-                if self.dbNames[name]
-                else f"Database {name} is not selected."
-            )
+    @pyqtSlot(str)
+    def setDB(self, name: str):
+        """Sets the database to fetch the numbers."""
```
kangz12345, line 129: Please write the docstring for `name`.

kangz12345 (changes requested):

examples/datacalc.py
BECATRUE: Wow.. I've been trying so hard to solve this. Thanks for the advice😂

examples/datacalc.py
BECATRUE: In `__init__()`, the initial value of `dbBox` is `""`, an empty string, which represents that no db is selected.

So, if the selected db is removed, it should be updated `""`. Otherswise, it will be reselected as an unintended db.

examples/datacalc.py
```diff
@@ -94,38 +99,42 @@ def updateDB(self, busName: str, msg: str):
...
                     if all(key in db for key in ("name", "path")):
                         name, path = db["name"], db["path"]
-                        self.dbs[name] = path
-                        for dbBox in self.viewerFrame.dbBoxes.values():
-                            dbBox.addItem(name)
+                        newDBs.add(name)
+                        if name not in self.dbs:
+                            self.dbs[name] = path
+                            for dbBox in self.viewerFrame.dbBoxes.values():
+                                dbBox.addItem(name)
```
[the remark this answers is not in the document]
BECATRUE: `map()` returns just an iterator, so that I wrapped it with `list()` and all iterators run.

BECATRUE:
I applied reviews.

examples/datacalc.py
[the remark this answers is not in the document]
kangz12345: Oh, you are right! My mistake.
In that case, your original version looks better 😄

kangz12345 (changes requested):

examples/datacalc.py
```diff
@@ -94,38 +101,48 @@ def updateDB(self, channelName: str, msg: str):
             except json.JSONDecodeError as e:
                 print(f"apps.datacalc.updateDB(): {e!r}")
             else:
```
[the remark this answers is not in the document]
BECATRUE: Why isn't the `else` statement necessary?
if `json.JSONDecodeError` occurs, we must not run the following code.

BECATRUE:
I applied the reviews.

examples/datacalc.py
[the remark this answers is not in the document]
kangz12345: You can just `return` in the `except:` block!

kangz12345 (changes requested):
Not necessary, but just for reducing nesting.

BECATRUE:
Ok! I appied it again.

kangz12345 (approved):
👍

PR #127, opened 2023-03-31
Becatrue/118/change bus to channel

### LINKED ISSUE #118: Change the word "bus" -> "channel"
How about changing the term "bus" to "channel"?
Since now there is no concept of actual buses, so it sounds a little awkward.
_Broadcasting_ to a _channel_ would be natural and intuitive.

BECATRUE:
This PR will close #118.

I changed all `bus`es to `channel`s, and set channel names as `dbch` and logch`.

### DIFF
--- a/swift/app.py
+++ b/swift/app.py
@@ -13,10 +13,10 @@ class BaseApp(QObject):
     """Base App class that all apps should inherit.
 
     Signals: 
-        broadcastRequested(str, str): A signal for broadcasting to a bus 
-          which contains the destination bus name and the message.
-        received(str, str): A signal for receiving a global signal from a bus
-          which contains the departure bus name and the message.
+        broadcastRequested(str, str): A signal for broadcasting to a channel 
+          which contains the destination channel name and the message.
+        received(str, str): A signal for receiving a global signal from a channel
+          which contains the departure channel name and the message.
     """
 
     broadcastRequested = pyqtSignal(str, str)
--- a/swift/app.py
+++ b/swift/app.py
@@ -14,9 +14,9 @@ class BaseApp(QObject):
 
     Signals: 
         broadcastRequested(str, str): A signal for broadcasting to a channel 
-          which contains the destination channel name and the message.
+          which contains the target channel name and the message.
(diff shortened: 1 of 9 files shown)


swift/app.py
```diff
@@ -13,10 +13,10 @@ class BaseApp(QObject):
     """Base App class that all apps should inherit.
 
     Signals: 
-        broadcastRequested(str, str): A signal for broadcasting to a bus 
-          which contains the destination bus name and the message.
-        received(str, str): A signal for receiving a global signal from a bus
-          which contains the departure bus name and the message.
+        broadcastRequested(str, str): A signal for broadcasting to a channel 
+          which contains the destination channel name and the message.
```
kangz12345, line 17: Now the expression "destination" becomes awkward. How about "the target channel"?

swift/app.py
```diff
@@ -13,10 +13,10 @@ class BaseApp(QObject):
...
+        received(str, str): A signal for receiving a global signal from a channel
+          which contains the departure channel name and the message.
```
kangz12345, line 19: How about "the source channel"?

swift/swift.py
```diff
@@ -41,7 +41,7 @@ class AppInfo:
         pos: Position on the main window; refer to Qt.DockWidgetArea enum.
           Should be one of "left", "right", "top", or "bottom", case-sensitive.
           Otherwise, defaults to Qt.LeftDockWidgetArea.
-        bus: Buses which the app subscribes to.
+        channel: Channeles which the app subscribes to.
```
kangz12345, line 44: May be you have simply replaced the words 😄
Please change it to "Channels".

kangz12345 (changes requested):

swift/swift.py
BECATRUE: Ah.. I checked the other files, but I don't think I can do it here..😂

kangz12345: I think the other "channels" are fine!
Channeles -> Channels

BECATRUE:
I updated it. @kangz12345

kangz12345 (approved):
LGTM

PR #130, opened 2023-04-04
Fix docstring in swift.swift.parse() about indentation

### LINKED ISSUE #128: [REFACTOR] Modify docstring in swift.swift
In `swift.swift.parse()`, the multiline docstring is indented in four spaces.

Let's modify it to 2 spaces indentation to fit our convention.

```python
"""Returns a new cls instance from a JSON string.

This is a convenience function for just unpacking the JSON string and gives them
as keyword arguments of the constructor of cls.
    
Args:
    cls: A class object.
    kwargs: A JSON string of a dictionary that contains the keyword arguments of cls.
        Positional arguments should be given with the argument names, just like
        the other keyword arguments.
        There must not exist arguments which are not in cls constructor.
"""
```

BECATRUE:
This PR will close #128.

### DIFF
--- a/swift/swift.py
+++ b/swift/swift.py
@@ -65,9 +65,9 @@ def parse(cls: Type[T], kwargs: str) -> T:
     Args:
         cls: A class object.
         kwargs: A JSON string of a dictionary that contains the keyword arguments of cls.
-            Positional arguments should be given with the argument names, just like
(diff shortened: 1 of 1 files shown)


kangz12345 (approved):
LGTM

PR #135, opened 2023-04-07
Simplify handling signals in apps

### LINKED ISSUE #133: [FEATURE] Simplify broadcasting in apps
Currently the apps emits signals and connects slots to broadcast or to receive broadcast messages.
This style is not so intuitive nor convenient.

Therefore, we want to provide an easy way to broadcast and receive messages.

Basically `BaseApp` class will be modified.

#### 1. Broadcasting
Implement a simple 'wrapper' method as follows:
```python
def broadcast(self, channel: str, content: Any):
  content = json.dumps(content)
  self.broadcastRequested.emit(channel, content)
```

#### 2. Receiving
Implement a pre-connected slot as follows:
```python
def receivedSlot(self, channel: str, content: Any):
  """This will be overridden by child classes."""

@pyqtSlot(str, str)
def _receivedMessage(self, channel: str, msg: str):
  """This is connected to self.received signal."""
  self.receivedSlot(channel, json.loads(msg))
```
A child class which inherits `BaseApp` will then override `receivedSlot()` to handle the received messages.
Note that the class can still implement its own slots for `received` signal.
We just provide a new convenient way _in addition_ to the previous method.

BECATRUE:
This PR will close #133.

I implemented handling signals simplily according to the discussions in #133.

### DIFF
--- a/swift/app.py
+++ b/swift/app.py
@@ -48,12 +48,11 @@ class BaseApp(QObject):
             content: Content to be broadcast.
         
         """
-        if not isinstance(content, str):
-            try:
-                content = json.dumps(content)
-            except TypeError as e:
-                print(f"swift.app.broadcast(): {e!r}")
-                return
+        try:
+            content = json.dumps(content)
+        except TypeError as e:
+            print(f"swift.app.broadcast(): {e!r}")
+            return
         self.broadcastRequested.emit(channelName, content)
 
 
--- a/swift/app.py
+++ b/swift/app.py
@@ -4,7 +4,8 @@ Base module for App.
 Every App class should be a subclass of BaseApp.
 """
 
-from typing import Optional, Iterable
+import json
+from typing import Any, Optional, Iterable
 
 from PyQt5.QtCore import QObject, pyqtSignal
 from PyQt5.QtWidgets import QWidget
@@ -38,3 +39,19 @@ class BaseApp(QObject):
             An iterable object of Frame objects for showing.
         """
         return ()
+
+    def broadcast(self, channelName: str, content: Any):
+        """Broadcasts the content to the target channel.
+
+        Args:
+            channelName: Target channel name.
+            content: Content to be broadcast.
+        
+        """
+        if not isinstance(content, str):
+            try:
+                content = json.dumps(content)
+            except TypeError as e:
+                print(f"swift.app.broadcast(): {e!r}")
+                return
+        self.broadcastRequested.emit(channelName, content)
--- a/swift/app.py
+++ b/swift/app.py
@@ -55,3 +55,17 @@ class BaseApp(QObject):
                 print(f"swift.app.broadcast(): {e!r}")
                 return
         self.broadcastRequested.emit(channelName, content)
+
+
+    def receivedSlot(self, channelName: str, content: Any):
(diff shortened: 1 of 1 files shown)


swift/app.py
```diff
@@ -38,3 +40,31 @@ def frames(self) -> Iterable[QWidget]:
             An iterable object of Frame objects for showing.
         """
         return ()
+
+    def broadcast(self, channelName: str, content: Any):
+        """Broadcasts the content to the target channel.
+
+        Args:
+            channelName: Target channel name.
+            content: Content to be broadcast.
```
kangz12345, line 49: Please specify that the `content` must be JSON-ify-able(?)

swift/app.py
```diff
@@ -38,3 +40,31 @@ def frames(self) -> Iterable[QWidget]:
...
+        """
+        try:
+            content = json.dumps(content)
+        except TypeError as e:
+            print(f"swift.app.broadcast(): {e!r}")
+            return
+        self.broadcastRequested.emit(channelName, content)
+
+    def receivedSlot(self, channelName: str, content: Any):
+        """This will be overridden by child classes."""
```
kangz12345, line 60: Please describe more about this slot, for example:
```python
"""Handles the received broadcast messages.

This is called when self.received signal is emitted.
The child class may override this method to implement a specific handler.
"""

File: swift/app.py [hunk 3]
```diff
@@ -38,3 +40,31 @@ def frames(self) -> Iterable[QWidget]:
...
+
+    @pyqtSlot(str, str)
+    def _receivedMessage(self, channelName: str, msg: str):
+        """This is connected to self.received signal."""
+        try:
+            msg = json.loads(msg)
+        except json.JSONDecodeError as e:
+            print(f"swift.app._receivedMessage(): {e!r}")
+            return
+        self.receivedSlot(channelName, msg)
```
kangz12345, line 70: This is just a preference, but in this case you may use `else:` block because it only has a single line (line 70).
I prefer to use `else:` block unless the block is too big.

However, you can keep this code, no problem!

kangz12345 (changes requested):

swift/app.py
BECATRUE: Ok! Also, I added arguments descriptions.

swift/app.py
BECATRUE: Oh, I agree with you. I will change it.

BECATRUE:
I updated it.

swift/app.py
```diff
@@ -38,3 +40,43 @@ def frames(self) -> Iterable[QWidget]:
...
+
+    def receivedSlot(self, channelName: str, content: Any):
+        """Handles the received broadcast message.
+        
+        This is called when self.received signal is emitted.
+        This will be overridden by child classes.
+
+        Args:
+            channelName: Channel name that transferred the message.
+            content: Content to be broadcast.
```
kangz12345, line 66: Since this is `received` slot, "received content" or something like that would be more appropriate.
"to be broadcast" looks like it transmits the content, rather that receives it.

kangz12345 (changes requested):

BECATRUE:
I updated it!

kangz12345 (approved):
LGTM

PR #138, opened 2023-04-08
Becatrue/132/swift call window

### LINKED ISSUE #132: [FEATURE] Add a window which allows swift-call
We want to add a window which allows `swift-call` request when an app sends it.

TBD

BECATRUE:
This PR will close #132.

I implmented a window to allow `swift-call` requests.
When the `swift` receives a `swift-call` request, it opens a window which contains two buttons; "Ok" and "Cancel".
If you click "OK" button, the request will be operated, and if you click "Cancel" button or close the window, it will be rejected.

The example implementation can be checked in `BECATRUE/132/swift-call-window-test` branch.

### DIFF
--- a/swift/swift.py
+++ b/swift/swift.py
@@ -223,7 +223,7 @@ class Swift(QObject):
             name, info = args["name"], args["info"]
             reply = QMessageBox.warning(
                 None, 
-                "Swift-call",
+                "swift-call",
                 f"The app {self.sender().name} requests for creating an app {name}",
                 QMessageBox.Ok | QMessageBox.Cancel,
                 QMessageBox.Cancel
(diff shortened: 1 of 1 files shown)


swift/swift.py
```diff
@@ -220,13 +220,31 @@ def _swiftcall(self, msg: str):
                 print("The message was ignored because "
                       "args of the create action have no such key; name or info.")
                 return
-            self.createApp(args["name"], AppInfo(**args["info"]))
+            name, info = args["name"], args["info"]
+            reply = QMessageBox.warning(
+                None,
+                "swift-call",
+                f"The app {self.sender().name} requests for creating an app {name}",
```
kangz12345, line 227: I think the message should contain `info` in order for the user to determine it.

kangz12345 (changes requested):

swift/swift.py
kangz12345: However, maybe my next PR would resolve this because I am going to generalize the swiftcall mechanism.

kangz12345 (approved):
LGTM!

Issue #137, opened 2023-04-08
[BUG] Not working all original message transfer protocol


BECATRUE:
### Describe the bug
After we simplified the broadcast/receive protocol, all original examples became not working.

In original system, we _directly_ emit and receive messages as _string_.
Now, however, the `received` signal connects to `_receivedMessage()` slot and it tries to decode the message from _json string_.
In this procedure, a `JSONDecodeError` occurs.

### To Reproduce
All messages should be changed to go through `broadcast()` and `receivedSlot()`.

BECATRUE:
This issue is closed by #140.

PR #140, opened 2023-04-09
Fix the error not working all original message transfer protocol

BECATRUE:
Yet, this PR is not completed. Just related to #137.

The problem is all broadcast-receive examples are not working, because it is only encoded by `broadcast()`, not decoded.
So, each communication occurs an error like below.

img

To fix it, I replaced all direct `emit()` into `broadcast()` and changed the receiving part a little.

### DIFF
--- a/examples/logger.py
+++ b/examples/logger.py
@@ -94,7 +94,7 @@ class LoggerApp(BaseApp):
         """Adds a channel name and log message.
 
         Args:
-            msg: Log message
+            content: Received log message.
         """
         timeString = time.strftime("%Y-%m-%d %H:%M:%S", time.localtime(time.time()))
         self.loggerFrame.logEdit.insertPlainText(f"{timeString}: {content}\n")
--- a/examples/logger.py
+++ b/examples/logger.py
@@ -3,7 +3,7 @@ App module for logging.
 """
 
 import time
-from typing import Optional, Tuple
+from typing import Any, Optional, Tuple
 
 from PyQt5.QtCore import QObject, pyqtSlot, pyqtSignal
 from PyQt5.QtWidgets import QWidget, QVBoxLayout, QPushButton, QTextEdit, QLabel, QDialogButtonBox
@@ -82,7 +82,6 @@ class LoggerApp(BaseApp):
         super().__init__(name, parent=parent)
         self.loggerFrame = LoggerFrame()
         # connect signals to slots
-        self.received.connect(self.addLog)
         self.loggerFrame.clearButton.clicked.connect(self.checkToClear)
         self.confirmFrame = ConfirmClearingFrame()
         self.confirmFrame.confirmed.connect(self.clearLog)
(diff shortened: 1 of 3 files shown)


BECATRUE:
First of all, I only modified `numgen`, `dbmgr`, and `logger`.
Please leave you opinion about these three apps!

examples/numgen.py
```diff
@@ -83,52 +82,62 @@ def __init__(self, name: str, table: str = "number", parent: Optional[QObject] =
...
+                self.generatorFrame.dbBox.addItem(name)
+        removingDBs = originalDBs - newDBs
+        if self.generatorFrame.dbBox.currentText() in removingDBs:
+            self.generatorFrame.dbBox.setCurrentText("")
+        for name in removingDBs:
+            self.dbs.pop(name)
+            self.generatorFrame.dbBox.removeItem(self.generatorFrame.dbBox.findText(name))
+
+    def receivedSlot(self, channelName: str, content: Any):
+        """Handles the received broadcast message.
```
kangz12345, line 124: Please use the notation agreed in #44.

examples/numgen.py
```diff
@@ -83,52 +82,62 @@ def __init__(self, name: str, table: str = "number", parent: Optional[QObject] =
...
 
-        This is a slot for received signal.
         It assumes that:
             The new database is always added at the end.
             Changing the order of the databases is not allowed.
 
         Args:
-            channelName: A name of the channel that transfered the signal.
-            msg: An input message to be transferred through the channel.
+            channelName: Channel name that transferred the message.
```
kangz12345, line 100: There is no argument `channelName`!

examples/logger.py
```diff
@@ -91,21 +90,41 @@ def frames(self) -> Tuple[LoggerFrame]:
...
         return (self.loggerFrame,)
 
-    @pyqtSlot(str, str)
-    def addLog(self, channelName: str, msg: str):
+    def addLog(self, content: str):
         """Adds a channel name and log message.
 
         Args:
-            channelName: The name of channel
             msg: Log message
```
kangz12345, line 97: Argument name mismatch: `content` vs. `msg`.

kangz12345 (changes requested):
As you saw here, we have to manually check the docstring and comments since our tools (e.g., vscode) cannot.
That's why we should try to write the code clearly enough and use comments only when it is necessary.
It would be hard to find outdated comments in the future 😄
Of course, in this PR, they ARE necessary docstrings though!

examples/numgen.py
BECATRUE: Oh, I forgot it. Thanks for letting me know.

BECATRUE:
I applied all reviews!

kangz12345 (approved):
LGTM

BECATRUE:
I completely forgot fixing other apps.. I will make another PR.

Issue #136, opened 2023-04-07
[FEATURE] Simplify swift-call request


kangz12345:
### Feature you want to implement
Just like the broadcasting in #133, I want to provide an easy way to request swift-calls.
This will be implemented right after #131 and #139 is implemented.

### How the feature is implemented
I want to implement a _proxy_ to request swift-call.
For example (usage):
```python
self.swiftcall.createApp(appinfos)
```
To receive the result, one may use a `Result` object as follows:
```python
result = self.swiftcall.createApp(appinfos)
...
if result.done:
  if result.success:
    use(result.value)
  else:
    raise result.error
```
However, I still have to think about the thread-safe implementation of the `Result` behavior.

PR #141, opened 2023-04-09
Swift-call result interface

### LINKED ISSUE #131: [FEATURE] System request result interface
We want to decide an interface of system request result to provide `swift` channel with a consistent form.

The system request result should contain the followings:
1. Requested feature (e.g., `createApp(...)`)
2. Requester app
3. Result status (e.g., success, failed, etc.)
4. Result details

kangz12345:
I implemented the interface for getting swift-call results.

Few insights:
1. The `Result` class will be used in apps to manage the swift-calls in the future. It will be implemented maybe in #136.
2. [`QObject.sender()`](https://doc.qt.io/qt-5/qobject.html#sender) is convenient, but we have to be careful as the document says, which implies that it easily breaks our assumptions and intuition. For example, consider the case where a slot is not called by a signal.
> Warning: This function violates the object-oriented principle of modularity.

3. Pylint complains about catching `Exception`. However, I intended to do that, so I applied [this solution](https://stackoverflow.com/a/64040518).
4. In a function docstring, we don't document the exceptions raised when the input violates the documented condition.

This closes #131.

### DIFF
--- a/swift/swift.py
+++ b/swift/swift.py
@@ -270,7 +270,7 @@ class Swift(QObject):
         sender = self.sender()
         try:
             value = self._handleSwiftcall(sender.name, msg)
-        except Exception as error:
+        except Exception as error: # pylint: disable=broad-exception-caught
             result = Result(done=True, success=False, error=repr(error))
         else:
             result = Result(done=True, success=True, value=value)
--- a/swift/swift.py
+++ b/swift/swift.py
@@ -247,7 +247,7 @@ class Swift(QObject):
             )
             if reply == QMessageBox.Ok:
                 return self.createApp(name, AppInfo(**info))
-            raise RuntimeError("user rejected the request.")
+            raise RuntimeError("The user rejected the request.")
         if action == "destroy":
             name = args["name"]
             reply = QMessageBox.warning(
@@ -259,7 +259,7 @@ class Swift(QObject):
             )
             if reply == QMessageBox.Ok:
                 return self.destroyApp(name)
-            raise RuntimeError("user rejected the request.")
+            raise RuntimeError("The user rejected the request.")
         raise NotImplementedError(action)
 
     def _swiftcall(self, sender: str, msg: str):
--- a/swift/swift.py
+++ b/swift/swift.py
@@ -223,6 +223,10 @@ class Swift(QObject):
               "destory": Destroy an app.
                 Its "args" is a dictionary with a key; "name".
                 The value of "name" is a name of app you want to destroy.
+        
+        Raises:
+            RuntimeError: When the user rejects the request.
+            NotImplementedError: When the given request action is not implemented.
         """
         msg = json.loads(msg)
         action, args = msg["action"], msg["args"]
@@ -237,6 +241,8 @@ class Swift(QObject):
             )
             if reply == QMessageBox.Ok:
                 self.createApp(name, AppInfo(**info))
+            else:
+                raise RuntimeError("user rejected the request.")
         elif action == "destroy":
             name = args["name"]
             reply = QMessageBox.warning(
@@ -248,9 +254,10 @@ class Swift(QObject):
             )
             if reply == QMessageBox.Ok:
                 self.destroyApp(name)
+            else:
+                raise RuntimeError("user rejected the request.")
         else:
-            print(f"The swift-call was ignored because "
-                  f"the treatment for the action {action} is not implemented.")
+            raise NotImplementedError(action)
 
 
 @contextmanager
--- a/swift/swift.py
+++ b/swift/swift.py
@@ -274,7 +274,7 @@ class Swift(QObject):
         """
         try:
             value = self._handleSwiftcall(sender, msg)
-        except Exception as error: # pylint: disable=broad-exception-caught
+        except Exception as error:  # pylint: disable=broad-exception-caught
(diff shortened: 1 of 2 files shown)


swift/swift.py
```diff
@@ -80,7 +81,23 @@ def strinfo(info: AppInfo) -> str:
...
+
+
+@dataclasses.dataclass
+class Result:
+    """Result data of a swift-call.
+    
+    Fields:
+        done: Whether the swift-call is done. Even when it failed, this is True as well.
+        success: True when the swift-call is done without any problems.
+        value: Return value of the swift-call, if any. It must be JSONifiable.
```
BECATRUE, line 94: The word "JSONifiable" looks fun😃

swift/swift.py
```diff
@@ -80,7 +81,23 @@ def strinfo(info: AppInfo) -> str:
     Args:
         info: Dataclass object to convert to a JSON string.
     """
-    return json.dumps(asdict(info))
+    return json.dumps(dataclasses.asdict(info))
+
+
+@dataclasses.dataclass
+class Result:
```
BECATRUE, line 88: I think just `Result` is so ambiguous. How about adding some explanations, such as `SwiftcallRequest`?

If you still don't mind, no problem!

swift/swift.py
```diff
@@ -204,50 +227,58 @@ def _swiftcall(self, msg: str):
...
+        Note that swiftcallRequested signal only has one str argument.
+        In fact the partial method will be connected using functools.partial().
+
+        Args:
+            sender: See _handleSwiftcall().
+            msg: See _handleSwiftcall().
+        """
+        try:
+            value = self._handleSwiftcall(sender, msg)
+        except Exception as error: # pylint: disable=broad-exception-caught
```
BECATRUE, line 277: Suppressing the warning looks good!

But, please start at least 2 spaces from the code. It is referred to [Google Python Style Guide](https://google.github.io/styleguide/pyguide.html#385-block-and-inline-comments).

swift/swift.py
```diff
@@ -204,50 +227,58 @@ def _swiftcall(self, msg: str):
...
                 QMessageBox.Ok | QMessageBox.Cancel,
                 QMessageBox.Cancel
             )
             if reply == QMessageBox.Ok:
-                self.destroyApp(name)
+                return self.destroyApp(name)
+            raise RuntimeError("user rejected the request.")
+        raise NotImplementedError(action)
+
+    def _swiftcall(self, sender: str, msg: str):
```
BECATRUE, line 265: Why don't you use `@pyqtSlot`?

I'm not sure, but if you use it, `sender` may be not necessary since it can be obtained by `self.sender()`.

swift/swift.py
```diff
@@ -204,50 +227,58 @@ def _swiftcall(self, msg: str):
...
-        try:
-            msg = json.loads(msg)
-        except json.JSONDecodeError as e:
-            print(f"swift.swift._swiftcall(): {e!r}")
-            return
-        if any(key not in msg for key in ("action", "args")):
-            print("The message was ignored because "
-                  "it has no such key; action or args.")
-            return
+        msg = json.loads(msg)
```
BECATRUE, line 237: Is it okay not to check if `msg` has a correct form?

swift/swift.py
```diff
@@ -204,50 +227,58 @@ def _swiftcall(self, msg: str):
...
                 "swift-call",
-                f"The app {self.sender().name} requests for destroying an app {name}",
+                f"The app {sender} requests for destroying an app {name}",
                 QMessageBox.Ok | QMessageBox.Cancel,
                 QMessageBox.Cancel
             )
             if reply == QMessageBox.Ok:
-                self.destroyApp(name)
+                return self.destroyApp(name)
+            raise RuntimeError("user rejected the request.")
```
BECATRUE, line 262: How about starting a capital letter?

BECATRUE (changes requested):
I left some reviews.

swift/swift.py
kangz12345: It makes sense, I will change it to `SwiftcallResult`!

swift/swift.py
kangz12345: As I said in the initial comment of this PR:

> 2. [QObject.sender()](https://doc.qt.io/qt-5/qobject.html#sender) is convenient, but we have to be careful as the document says, which implies that it easily breaks our assumptions and intuition. For example, consider the case where a slot is not called by a signal.

I would prefer not to use `self.sender()` if possible.

swift/swift.py
kangz12345: I wrote

> This can raise an exception if the arguments do not follow the valid API.
The caller must obey the API and catch the possible exceptions.

in the docstring of `_handleSwiftcall()`.

I understand that we agreed to catch all the exceptions as possible to prevent the application to crash.
However, this is only called by `_swiftcall()` and it catches all the exceptions.
Moreover, this is an internal method, i.e., will not be used in other places.
I think this usage is safe enough!

The main reason of not catching the exceptions is that the exception information should be transferred to the caller app.

swift/swift.py
kangz12345: Okay, I agree with that!

BECATRUE (approved):
LGTM!

PR #142, opened 2023-04-12
Generalize swift call message protocol

### LINKED ISSUE #139: [FEATURE] Generalize swift-call message protocol
The current implementation of swift-call message protocol requires handling case by case.
Moreover, the message boxes introduced in #132 should be implemented for each _action_ when we add more swift-calls in the future.

Therefore, I want to generalize this protocol.

1. Swift-call messages follow the format below:
```python
{
  "call": "destroyApp", # the method name of Swift to call
  "args": {"name": "test-app"}, # arguments of the method, all as keyword arguments
}
```
2. The swift-call handler will try to execute the given method call.
3. If it fails with an exception, catch it and inform the caller app.

However, this has a limitation that some swift-calls' arguments cannot be converted to JSON strings.
In these cases, I will provide a way to take JSON strings as one of the following:
1. Define another method `createAppFromJSON(name: str, info: str)` that takes a JSON string and convert it to a dictionary and then to an `AppInfo` object. Then it will call the original `createApp()` with the `AppInfo` object.
2. Allow the original `createApp()` method to take a string as well, i.e., the type hint for `info` becomes `Union[AppInfo, str]`, and use `isinstance(info, str)`.

The first option can clearly separate the method for public API and for swiftcall.
The second option is convenient to implement.
Your opinions are welcome!

kangz12345:
I refactored `_handleSwiftcall()` to accept more general method calls.

Summary:

1. I introduced `Serializable` class to indicate the dataclasses that are able to converted to JSON strings. All the public methods should have parameters that are JSONifiable by itself or an instance of `Serializable`.
2. `_parseArgs()` inspects the function signature of the public API methods, and converts arguments from JSON strings to concrete dataclass instances in the parameter type is `Serializable`.

However, I have not tested the actual behavior yet...

This closes #139.

### DIFF
--- a/swift/swift.py
+++ b/swift/swift.py
@@ -106,6 +106,9 @@ class SwiftcallInfo(Serializable):
           This is case-sensitive.
         args: The arguments of the swift-call as a dictionary of keyword arguements.
           The names of the arguements are case-sensitive.
+          When an argument is Serializable, it must be given as a converted JSON string,
+          e.g., not {"arg": SwiftcallInfo(call="call")},
+          but {"arg": '{"call": "call", "args": {}}'}.
     """
     call: str
     args: Mapping[str] = dataclasses.field(default_factory=dict)
--- a/swift/swift.py
+++ b/swift/swift.py
@@ -89,15 +89,13 @@ def parse(cls: Type[T], kwargs: str) -> T:
 
 # TODO(kangz12345): Method name looks too specific.
 # Change this to dumps() whenever we can rename this symbol.
-def strinfo(info: Serializable) -> str:
-    """Returns a JSON string converted from the given info.
-
-    This is just a convenience function for users not to import dataclasses and json.
+def strinfo(obj: Serializable) -> str:
+    """Returns a JSON string converted from the given Serializable object.
     
     Args:
-        info: Dataclass object to convert to a JSON string.
+        obj: Dataclass object to convert to a JSON string.
     """
-    return json.dumps(dataclasses.asdict(info))
+    return json.dumps(dataclasses.asdict(obj))
 
 
 @dataclasses.dataclass
--- a/swift/swift.py
+++ b/swift/swift.py
@@ -15,12 +15,13 @@ import argparse
 import json
 import importlib
 import importlib.util
+import inspect
 import dataclasses
 import functools
 from collections import defaultdict
 from contextlib import contextmanager
 from typing import (
-    Dict, Any, Iterable, Mapping, Optional, TypeVar, Type
+    Dict, Any, Callable, Iterable, Mapping, Optional, TypeVar, Type
 )
 
 from PyQt5.QtCore import QObject, pyqtSlot, Qt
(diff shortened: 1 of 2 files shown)


swift/swift.py
```diff
@@ -73,7 +87,9 @@ def parse(cls: Type[T], kwargs: str) -> T:
     return cls(**json.loads(kwargs))
 
 
-def strinfo(info: AppInfo) -> str:
+# TODO(kangz12345): Method name looks too specific.
+# Change this to dumps() whenever we can rename this symbol.
+def strinfo(info: Serializable) -> str:
     """Returns a JSON string converted from the given info.
 
     This is just a convenience function for users not to import dataclasses and json.
```
BECATRUE, line 95: Now we use this function. so the docstring should be updated.

swift/swift.py
```diff
@@ -217,50 +275,31 @@ def _handleSwiftcall(self, sender: str, msg: str) -> Any:
...
-                f"The app {sender} requests for destroying an app {name}",
-                QMessageBox.Ok | QMessageBox.Cancel,
-                QMessageBox.Cancel
-            )
-            if reply == QMessageBox.Ok:
-                return self.destroyApp(name)
-            raise RuntimeError("The user rejected the request.")
-        raise NotImplementedError(action)
+        info = parse(SwiftcallInfo, msg)
+        if info.call.startswith("_"):
```
BECATRUE, line 289: This restriction looks cool!

BECATRUE (changes requested):
Your code is very neat😃

I just left a few reviews.

swift/swift.py
```diff
@@ -57,6 +69,8 @@ class AppInfo:
     args: Optional[Mapping[str, Any]] = None
 
 
+# TODO(kangz12345): Make this and strinfo() a pair.
+# Change this to loads() whenever we can rename this symbol.
```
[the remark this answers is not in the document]
kangz12345: Okay, then I will change those in this PR!

swift/swift.py
kangz12345: Great point!

swift/swift.py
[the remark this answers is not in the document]
kangz12345: I renamed the symbols using VSCode. It seems it only changes the testing codes.

swift/swift.py
```diff
@@ -57,7 +69,7 @@ class AppInfo:
     args: Optional[Mapping[str, Any]] = None
 
 
-def parse(cls: Type[T], kwargs: str) -> T:
+def loads(cls: Type[T], kwargs: str) -> T:
     """Returns a new cls instance from a JSON string.
 
     This is a convenience function for just unpacking the JSON string and gives them
```
BECATRUE, line 75: Please remove this, too!

BECATRUE (changes requested):
Just a comment.

BECATRUE (approved):
LGTM!

PR #147, opened 2023-04-16
Kangz12345/136/simplify swift call request

### LINKED ISSUE #136: [FEATURE] Simplify swift-call request

kangz12345:
I implemented the simplified swift-call interface.
The newly introduced usage is as follows:
1. Every `BaseApp` has a new attribute: `swiftcall`.
2. From that, one can _call_ any swift-call just like calling a method, e.g., `self.swiftcall.destroyApp(name="logger")`. All the arguments must be given as keyword arguments, for clarity.
3. It will return a `SwiftcallResult` instance. It is updated when the swift-call is done and the result comes back via the `swiftcallReturned` signal. One may want to keep it to check the result, e.g.:
```python
result = self.swiftcall.destroyApp(name="logger")
...
if result.done:
  if result.success:
    print(f"returned value is {result.value}.")
  else:
    print(f"failed: {result.error}.")
```

4. One must not _block_ on the result to be done, e.g., `while not result.done: ...`, in the main thread, since it will freeze the entire GUI process.

I have tested `createApp()` and `destroyApp()`.
As I test it, I fixed a bug in `swift` module that I implemented in #142.

This closes #136.

### DIFF
--- a/swift/app.py
+++ b/swift/app.py
@@ -91,7 +91,7 @@ class BaseApp(QObject):
             self.receivedSlot(channelName, content)
 
 
-class SwiftcallProxy:
+class SwiftcallProxy:  # pylint: disable=too-few-public-methods
     """A proxy for requesting swift-calls conveniently.
     
     Every attribute access is proxied, and if you try to call a method of this
--- a/swift/app.py
+++ b/swift/app.py
@@ -10,6 +10,8 @@ from typing import Any, Optional, Callable, Iterable
 from PyQt5.QtCore import QObject, pyqtSignal, pyqtSlot
 from PyQt5.QtWidgets import QWidget
 
+from swift import swift
+
 class BaseApp(QObject):
     """Base App class that all apps should inherit.
 
@@ -108,6 +110,7 @@ class SwiftcallProxy:
         """
         self.requested = requested
         self.returned = returned
+        self.results = {}
 
     def __getattribute__(self, call: str) -> Callable:
         """Returns a callable object which emits a swift-call requesting signal.
@@ -115,3 +118,28 @@ class SwiftcallProxy:
         Args:
             call: The name of the swift-call.
         """
+        def proxy(**args: Any) -> swift.SwiftcallResult:
+            """Emits a swift-call request signal with the given arguments.
+
+            It saves the returned result to self.results dictionary, so when
+            self.returned signal is emitted, i.e., the swift-call result is received,
+            it will update the result object contents.
+
+            Args:
+                **args: The arguments for the swift-call, all as keyword arguments.
+                  If an argument is a swift.Serializable instance, it will be
+                  converted to a JSON string by swift.dumps().
+
+            Returns:
+                A swift-call result object to keep tracking the result.
+            """
+            for name in args:
+                if isinstance(args[name], swift.Serializable):
+                    args[name] = swift.dumps(args[name])
+            info = swift.SwiftcallInfo(call=call, args=args)
+            result = swift.SwiftcallResult(done=False, success=False)
+            msg = swift.dumps(info)
+            self.results[msg] = result
+            self.requested.emit(msg)
+            return result
+        return proxy
--- a/swift/app.py
+++ b/swift/app.py
@@ -140,3 +140,19 @@ class SwiftcallProxy:  # pylint: disable=too-few-public-methods
             self.requested.emit(msg)
             return result
         return proxy
+
+    def update_result(self, request: str, result: swift.SwiftcallResult):
+        """Updates the result for the request parsing the received message.
+
+        Args:
+            request: The request message that has been sent to Swift.
+            result: The received result object.
+        """
+        _result = self.results.get(request, None)
+        if _result is None:
+            print(f"SwiftcallProxy._update_result: Failed to find a result for {request}.")
+            return
+        _result.error = result.error
+        _result.value = result.value
+        _result.success = result.success
+        _result.done = result.done
--- a/swift/app.py
+++ b/swift/app.py
@@ -158,6 +158,8 @@ class SwiftcallProxy:  # pylint: disable=too-few-public-methods
             info = swift.SwiftcallInfo(call=call, args=args)
             result = swift.SwiftcallResult(done=False, success=False)
             msg = swift.dumps(info)
+            if msg in self.results:
+                print(f"SwiftcallProxy.<local>.proxy(): Duplicate message {msg} is ignored.")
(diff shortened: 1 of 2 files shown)


BECATRUE:
First of all, I wonder how you tested `createdApp()` and `destroyApp()`.
Concretely, how did you check if `result` is updated not using `while not result.done: ...`?

swift/swift.py
```diff
@@ -255,7 +255,7 @@ def _parseArgs(self, call: Callable, args: Mapping[str, Any]) -> Dict[str, Any]:
         parsedArgs = {}
         for name, arg in args.items():
             cls = signature.parameters[name].annotation
-            parsedArgs[name] = loads(cls, arg) if isinstance(cls, Serializable) else arg
+            parsedArgs[name] = loads(cls, arg) if issubclass(cls, Serializable) else arg
```
BECATRUE, line 258: Oh, it was a very difficult error to find😂

kangz12345:
> First of all, I wonder how you tested `createdApp()` and `destroyApp()`.
> Concretely, how did you check if `result` is updated not using `while not result.done: ...`?

I'm sorry for the missing test codes... 😄 
I used the interactive console for testing.
So I import `swift` and load the setup.json file, and then I call the swift-calls. Then I click the OK button from the prompt, and finally I check if the result object is updated.

swift/app.py
```diff
@@ -5,21 +5,30 @@
...
     """Base App class that all apps should inherit.
 
     Signals: 
-        broadcastRequested(str, str): A signal for broadcasting to a channel 
-          which contains the target channel name and the message.
-        received(str, str): A signal for receiving a global signal from a channel
-          which contains the source channel name and the message.
-        TODO(kangz12345): docstring for swiftcallRequested and swiftcallReturned.
-          They should be written when the swiftcall protocol is finally determined.
+        broadcastRequested(str, str): The app requested broadcasting to a channel
```
BECATRUE, line 19: It seems that the past expression can cause a little confusion.
How about including that this signal will be emitted once the app requests broadcasting?

swift/app.py
```diff
@@ -35,7 +44,9 @@ def __init__(self, name: str, parent: Optional[QObject] = None):
         """
         super().__init__(parent=parent)
         self.name = name
+        self.swiftcall = SwiftcallProxy(self.swiftcallRequested)
         self.received.connect(self._receivedMessage)
+        self.swiftcallReturned.connect(self._receivedResult)
```
BECATRUE, line 49: How about including 'swiftcall' in the name of slot?
So far, we included 'swiftcall' in names about `swiftcall`, such as `swiftcallReqeusted`, `swiftcall`, etc.

swift/app.py
```diff
@@ -84,3 +95,86 @@ def _receivedMessage(self, channelName: str, msg: str):
...
+    """
+
+    def __init__(self, requested: QObject):
+        """
+        Args:
+            requested: A pyqtSignal(str) which will be emitted when a proxied
+              method call is invoked. See BaseApp.swiftcallRequested.
+        """
+        self.requested = requested
+        self.results: Dict[str, swift.SwiftcallResult] = {}
```
BECATRUE, line 132: It looks good to distinguish various reqeusts from their message (`SwiftcallInfo`).
However, I'm worried a little because some requests may be the same.
(In current system, it will NOT occur, but someday..)

What do you think of it? If you're okay, it doesn't matter!

swift/app.py
```diff
@@ -84,3 +95,86 @@ def _receivedMessage(self, channelName: str, msg: str):
...
+    def update_result(self, request: str, result: swift.SwiftcallResult):
+        """Updates the result for the request parsing the received message.
+
+        Args:
+            request: The request message that has been sent to Swift.
+            result: The received result object.
+        """
+        _result = self.results.get(request, None)
+        if _result is None:
+            print(f"SwiftcallProxy._update_result: Failed to find a result for {request}.")
```
BECATRUE, line 175: It is just a coding style, but how about modifying `SwiftcallProxy._update_result()`.
So far, we have adopted this form.

BECATRUE (changes requested):
The proxy idea looks excellent!

I left a few reviews.

swift/app.py
kangz12345: Okay, I will make the expression more precise.

swift/app.py
kangz12345: Good idea! I will change the name.

swift/app.py
kangz12345: I thought about that as well, but I wanted to keep the mechanism simple as possible.
How about tackling that when it becomes an actual problem?

swift/app.py
kangz12345: Oh, I didn't know that. Thanks!
Moreover I will change that `.update_result()`, as the method is public.

swift/app.py
BECATRUE: I agree with you. I like the simple mechanism now, too!

BECATRUE: Then, please implement checking if the key already exists.

kangz12345:
I applied the reviews and add one more feature in [the last commit](https://github.com/snu-quiqcl/swift/pull/147/commits/b46c581d330aa10282d633fb3a2c2444398f398b).

swift/app.py
```diff
@@ -84,3 +95,93 @@ def _receivedMessage(self, channelName: str, msg: str):
...
+    def update_result(self, request: str, result: swift.SwiftcallResult, discard: bool = True):
+        """Updates the result for the request parsing the received message.
+
+        Args:
+            request: The request message that has been sent to Swift.
+            result: The received result object.
+            discard: If True, the result object is removed from self.results.
+              In most cases, it will be updated only once and never be looked up again.
+              Therefore, it is efficient to discard it after updating the result.
+              If you want to find the result from self.results later again, give False.
```
BECATRUE, line 177: The idea to discard updated results looks good!

BECATRUE (approved):
Perfect!

PR #150, opened 2023-04-20
Implement swift-call 'updateFrames()' which updates frames of the app

### LINKED ISSUE #149: [FEATURE] Implement swift-call 'updateFrames()' which updates frames of the app
### Discussed in https://github.com/snu-quiqcl/swift/discussions/148

<sup>Originally posted by **BECATRUE** April 18, 2023</sup>
In current system, we can craate a new app using `self.swiftcall.createApp(...)`.
However, we cannot create a new frame in an existing app.

For example, there is an app `APP1` and its frames are `FRAME1` and `FRAME2`, but `APP1.frames()` returns only `FRAME1`.
In this case, an user cannot open `FRAME2` at any time.

So, how about adding swift-call `createFrame`?</div>

BECATRUE:
This PR will close #149.

I implemented a swift-call `updateFrames()` which updates frames of the target app.

It has two arguments:
- name: the target app name
- update: If true, it will remove all original frames and add all new frames.
Otherwise, it will remove only 'original frames - new frames' and add only 'new frames - original frames', i.e. remain the intersection of original frames and new frames.

### DIFF
--- a/swift/swift.py
+++ b/swift/swift.py
@@ -233,12 +233,11 @@ class Swift(QObject):
             apps.discard(app)
         app.deleteLater()
 
-    def updateFrames(self, name: str, update: bool = False):
+    def updateFrames(self, name: str):
         """Updates the frames of an app.
         
         Args:
             name: A name of the app to update its frames.
-            update: Whether the original frames will be updated or remained.
         """
         app = self._apps[name]
         info = self.appInfos[name]
@@ -246,11 +245,11 @@ class Swift(QObject):
         newFrames = app.frames()
         orgFramesSet = set(orgFrames)
         newFramesSet = set(newFrames)
-        for frame in orgFramesSet if update else orgFramesSet - newFramesSet:
+        for frame in orgFramesSet - newFramesSet:
             dockWidget = orgFrames[frame]
             self.mainWindow.removeDockWidget(dockWidget)
             dockWidget.deleteLater()
-        for frame in newFramesSet if update else newFramesSet - orgFramesSet:
+        for frame in newFramesSet - orgFramesSet:
             self.addFrame(name, frame, info)
 
     @pyqtSlot(str, str)
--- a/swift/swift.py
+++ b/swift/swift.py
@@ -148,6 +148,7 @@ class Swift(QObject):
             parent: A parent object.
         """
         super().__init__(parent=parent)
+        self.appInfos = appInfos
         self.mainWindow = QMainWindow()
         self.centralWidget = QLabel("Swift")
         self.centralWidget.setAlignment(Qt.AlignCenter)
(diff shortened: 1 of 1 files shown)


kangz12345:
When do we need the `update` feature?

swift/swift.py
```diff
@@ -222,6 +223,36 @@ def destroyApp(self, name: str):
...
+            dockWidget.setWidget(frame)
+            area = {
+                "left": Qt.LeftDockWidgetArea,
+                "right": Qt.RightDockWidgetArea,
+                "top": Qt.TopDockWidgetArea,
+                "bottom": Qt.BottomDockWidgetArea
+            }.get(info.pos, Qt.LeftDockWidgetArea)
+            if info.show:
+                self.mainWindow.addDockWidget(area, dockWidget)
+            self._dockWidgets[name].append(dockWidget)
```
kangz12345, line 254: Can we make this a separate method?

kangz12345 (changes requested):
A question emerges now: why did we give the `show` option for each app, not frame? 😅

BECATRUE:
> A question emerges now: why did we give the `show` option for each app, not frame? 😅

I have the same question, too..😂
However, we cannot determine how many frames will be returned by `frames()`, so I think using one unified `show` is the best.

BECATRUE:
> When do we need the `update` feature?

Umm.. It's just a abstract situation, but I think that we may want to reopen a frame for re-initializing.
If you think it is not necessary, I will remove it.

kangz12345:
> Umm.. It's just a abstract situation, but I think that we may want to reopen a frame for re-initializing. If you think it is not necessary, I will remove it.

However, if we re-initialize (hence re-instantiate) a frame, then the instance is changed, i.e., it is _replaced_ with a different frame.
Therefore, the previous frame widget instance will be removed and the new instance be added even when `update=True`, I guess..!

BECATRUE:
> > Umm.. It's just a abstract situation, but I think that we may want to reopen a frame for re-initializing. If you think it is not necessary, I will remove it.
> 
> However, if we re-initialize (hence re-instantiate) a frame, then the instance is changed, i.e., it is _replaced_ with a different frame. Therefore, the previous frame widget instance will be removed and the new instance be added even when `update=True`, I guess..!

Oh, you're right.. I will remove it🥲

BECATRUE:
I applied the reviews and push the test branch `BECATRUE/149/update-frames-test`.
In the test, when started, in `numgen`, there is only one frame `generatorFrame`.
Then, if you generate a number, `viewerFrame` will be opened.

swift/swift.py
```diff
@@ -173,6 +173,26 @@ def load(self, appInfos: Mapping[str, AppInfo]):
         for name, info in appInfos.items():
             self.createApp(name, info)
 
+    def addFrame(self, name: str, frame: QWidget, info: AppInfo):
+        """Adds a frame of the app and wraps it with a dock widget.
+
```
kangz12345, line 178: It would be better to note:
`This is not a swift-call because QWidget is not Serializable.`

swift/swift.py
```diff
@@ -223,35 +233,24 @@ def destroyApp(self, name: str):
...
         info = self.appInfos[name]
         orgFrames = {dockWidget.widget(): dockWidget for dockWidget in self._dockWidgets[name]}
         newFrames = app.frames()
         orgFramesSet = set(orgFrames)
         newFramesSet = set(newFrames)
-        for frame in orgFramesSet if update else orgFramesSet - newFramesSet:
+        for frame in orgFramesSet - newFramesSet:
             dockWidget = orgFrames[frame]
             self.mainWindow.removeDockWidget(dockWidget)
             dockWidget.deleteLater()
```
kangz12345: Oops, please make this a separate method, e.g., `removeFrame()`, as it its usage is the same as in `destroyApp()`.

kangz12345 (changes requested):
Final review!

swift/swift.py
BECATRUE: Oh, the removing part is so short that I couldn't notice it.

BECATRUE:
I applied the reviews, again! Plz check it.

kangz12345:
> I applied the reviews, again! Plz check it.

I think you didn't push your work! @BECATRUE

BECATRUE:
> > I applied the reviews, again! Plz check it.
> 
> I think you didn't push your work! @BECATRUE

I updated now.. Sorry🥲

kangz12345 (approved):
LGTM

PR #158, opened 2023-04-28
Implement several examples for swift-call

### LINKED ISSUE #134: [FEATURE] Implement several examples for swift-call
In order for beginners to understand `swift-call`, we need several examples to show `swift-call`.
For this, I want to implement three examples; create / destroy an app, create a frame.

#### Create an app
- Remove `datacalc` from `setup.json`.
- Add a button in `dbmgr`, which creates `datacalc` when clicked.

#### Destroy an app
- If `datacalc` exists, the above button destroys `datacalc` when clicked.

#### Create a frame
- In initial setting, `frames()` of `numgen` returns only `generatorFrame`, not `viewerFrame`.
- One a number is generated, create the frame `viewerFrame`.

BECATRUE:
This PR will close #134.

In `dbmgr.openCloseDatacalc()`, there is no case when `self.openCloseDatacalcResult.done` is `False` because the GUI is freezing until the permission window closes, but to clarify, I checked it.

### DIFF
--- a/examples/dbmgr.py
+++ b/examples/dbmgr.py
@@ -98,6 +98,7 @@ class DBMgrApp(BaseApp):
         """Extended."""
         super().__init__(name, parent=parent)
         self.dbList = []
+        self.isDatacalcOpen = False
         self.openCloseDatacalcResult = None
         self.managerFrame = ManagerFrame()
         # connect signals to slots
@@ -170,19 +171,26 @@ class DBMgrApp(BaseApp):
 
     @pyqtSlot()
     def openCloseDatacalc(self):
-        self.openCloseDatacalcResult = self.swiftcall.createApp(
-            name="datacalc",
-            info=AppInfo(
-                module="examples.datacalc",
-                cls="DataCalcApp",
-                show=True,
-                pos="top",
-                channel=["db"],
-                args={
-                    "tables": {
-                        "A": "number",
-                        "B": "B"
+        if self.openCloseDatacalcResult:
+            if not self.openCloseDatacalcResult.done:
+                print("DBMgrApp.openCloseDatacalc(): The previous swiftcall must be done.")
+                return
+            if self.openCloseDatacalcResult.success:
+                self.isDatacalcOpen = not self.isDatacalcOpen
+        else:
+            self.openCloseDatacalcResult = self.swiftcall.createApp(
+                name="datacalc",
+                info=AppInfo(
+                    module="examples.datacalc",
+                    cls="DataCalcApp",
+                    show=True,
+                    pos="top",
+                    channel=["db"],
+                    args={
+                        "tables": {
+                            "A": "number",
+                            "B": "B"
+                        }
                     }
-                }
+                )
             )
-        )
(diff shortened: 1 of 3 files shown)


examples/dbmgr.py
```diff
@@ -162,3 +169,36 @@ def removeDB(self):
...
         self.sendDB(False, db.name)
+
+    @pyqtSlot()
+    def openCloseDatacalc(self):
+        """Opens or closes a datacalc app.
+
+        If the dataclac app is open, close it.
+        Otherwise, open a new datacalc app.
+        """
+        if self.openCloseDatacalcResult:
```
kangz12345, line 180: Please make it explicit: `... is not None`, even if we have not defined the boolean value of the result objects.

examples/dbmgr.py
```diff
@@ -162,3 +169,36 @@ def removeDB(self):
...
+            if not self.openCloseDatacalcResult.done:
+                print("DBMgrApp.openCloseDatacalc(): The previous swiftcall must be done.")
+                return
+            if self.openCloseDatacalcResult.success:
+                self.isDatacalcOpen = not self.isDatacalcOpen
+        if self.isDatacalcOpen:
+            self.openCloseDatacalcResult = self.swiftcall.destroyApp(name="datacalc")
```
kangz12345, line 187: Maybe the button text should be changed.

examples/numgen.py
```diff
@@ -87,7 +88,8 @@ def __init__(self, name: str, table: str = "number", parent: Optional[QObject] =
 
     def frames(self) -> Tuple[GeneratorFrame, ViewerFrame]:
         """Overridden."""
-        return (self.generatorFrame, self.viewerFrame)
+        return (self.generatorFrame, self.viewerFrame) if self.isGenerated \
```
kangz12345, line 91: In this case of a long line, use the `if-else` statement.

kangz12345 (changes requested):

examples/dbmgr.py
BECATRUE: It looks more clear!

examples/dbmgr.py
BECATRUE: Wow.. I completely forgot🥲 I will change it.

BECATRUE: Umm.. I need to track the result to set the text to show exactly, but I didn't want to implement it because I was only trying to show a simple example. Can I just show 'open or close datacalc'? @kangz12345

BECATRUE:
I applied the reviews!

examples/numgen.py
```diff
@@ -86,8 +87,14 @@ def __init__(self, name: str, table: str = "number", parent: Optional[QObject] =
         self.generatorFrame.generateButton.clicked.connect(self.generateNumber)
 
     def frames(self) -> Tuple[GeneratorFrame, ViewerFrame]:
```
kangz12345, line 89: Now this type hinting becomes incorrect.
The correct type would be:
```python
Union[Tuple[GeneratorFrame, ViewerFrame], Tuple[GeneratorFrame]]
```

kangz12345 (changes requested):

examples/numgen.py
BECATRUE: I applied it, again!

kangz12345 (approved):
LGTM

PR #159, opened 2023-04-28
Make `swift` be a flat module, not a package

### LINKED ISSUE #153: Merge `swift` and `app` modules and make `swift` as a flat module, not a package
I found that the two modules, `swift` and `app`, are in fact dependent to each other.
That implies that they should be in a single module, or we separated something not properly.
I realized this because of a pylint warning `cyclic-import`.
As far as I understand, a module is an independent group of codes and dependences should be acyclic.

For example, in `swift` module, a `Swift` instance manages several `BaseApp` instances where `BaseApp` is defined in `app` module.
However, in `app` module, it uses some functions and classes in `swift` module.
I think it would be better to put them together in a single module.

How do you think of this? @BECATRUE

kangz12345:
The `app` module is just copied and pasted into the `swift` module, and the `swift` module is moved out of the folder, hence became a flat module.

Only the importing and namespaces (e.g., `app.BaseApp` -> `swift.BaseApp`) parts are modified.

This closes #153.


BECATRUE (approved):
Thanks for modifying even the test file!

PR #160, opened 2023-05-02
Fix the bug about the type of Swift._subscribers

BECATRUE:
I integrated the type of `self._subscribers` into `DefaultDict[str, Set[str]]`.


kangz12345 (approved):
LGTM

PR #161, opened 2023-05-02
Testing for updated parts

### LINKED ISSUE #144: [TEST] Testing for updated parts
After all issues, we need to test for updated parts. This is the last order.

BECATRUE:
This will close #144.

In this issue, I tested all classes and functions in `swift` module.

For each class, it is implemented as each test class, such as `SwiftTest`.
In the above class, for each method is implemented as each method, such as `test_blabla()`.
I didn't implement tests for methods which are called when the class instance is initiated, e.g. `create_app()`.

And also, I realized a small bug in `subscribe()` in terms of type, so that I fixed it.

The coverage is 100%!


kangz12345:
Please summarize what did you do in this PR!

tests/test_swift.py
```diff
@@ -87,35 +73,139 @@ def setUp(self):
...
             self.assertNotIn(name, self.swift._apps)
 
+    def test_update_frames(self):
+        self.swift._apps["app1"].frames.return_value = (QWidget(),)
+        self.swift.updateFrames("app1")
+
+    def test_channel_names(self):
+        channelNames = self.swift.channelNames()
+        for channelName in self.channels:
+            self.assertIn(channelName, channelNames)
```
kangz12345, line 102: This also does not verify the equality.

tests/test_swift.py
```diff
@@ -87,35 +73,139 @@ def setUp(self):
...
+
+    def test_subscriber_names(self):
+        for channel in self.channels:
+            subscriberNames = self.swift.subscriberNames(channel)
+            for name, info in APP_INFOS.items():
+                if channel in info.channel:
+                    self.assertIn(name, subscriberNames)
+
```
kangz12345, line 110: Missing test for `subscribe()`.

tests/test_swift.py
```diff
@@ -87,35 +73,139 @@ def setUp(self):
...
+    def test_unsubcribe(self):
+        self.assertEqual(self.swift.unsubscribe("app1", "ch1"), True)
+        self.assertNotIn("app1", self.swift._subscribers["ch1"])
```
kangz12345, line 113: How about using the public API, i.e., `self.swift.subscriberNames("ch1")`?

tests/test_swift.py
```diff
@@ -87,35 +73,139 @@ def setUp(self):
...
+        self.assertEqual(self.swift.unsubscribe("app2", "ch1"), False)
+
+    def test_swiftcall(self):
+        QMessageBox.warning = MagicMock(return_value=QMessageBox.Ok)
```
kangz12345, line 117: Wow 😮

tests/test_swift.py
```diff
@@ -87,35 +73,139 @@ def setUp(self):
...
+        self.swift._swiftcall("app1", json.dumps({
+            "call": "createApp",
+            "args": {"name": "app3", "info": {"module": "module2", "cls": "cls2"}}
+        }))
+        self.swift.callWithSerializable = MagicMock()
+        self.swift.getSerializable = MagicMock(return_value=APP_INFOS["app1"])
+        self.swift._swiftcall("app1", json.dumps({"call": "getSerializable", "args": {}}))
+        self.swift._swiftcall("app1", json.dumps({"call": "_broadcast", "args": {}}))
+        QMessageBox.warning.return_value = QMessageBox.Cancel
+        self.swift._swiftcall("app1", json.dumps({"call": "appNames", "args": {}}))
```
kangz12345, line 128: Hmm.. maybe I should implement the test codes for swiftcalls since I implemented it. Let's just leave this codes to keep the 100% coverage.
I will implemented the test codes after this PR.

tests/test_swift.py
```diff
@@ -87,35 +73,139 @@ def setUp(self):
...
+    def setUp(self):
+        self.app = swift.BaseApp("name")
+
+    def test_init(self):
+        self.assertEqual(self.app.name, "name")
+        self.assertIsInstance(self.app.swiftcall, swift.SwiftcallProxy)
+        self.assertEqual(type(self.app.broadcastRequested), pyqtBoundSignal)
+        self.assertEqual(type(self.app.received), pyqtBoundSignal)
+        self.assertEqual(type(self.app.swiftcallRequested), pyqtBoundSignal)
+        self.assertEqual(type(self.app.swiftcallReturned), pyqtBoundSignal)
```
kangz12345, line 149: Is this necessary..? Is it for the coverage?

tests/test_swift.py
```diff
@@ -87,35 +73,139 @@ def setUp(self):
...
+
+    def test_set_parent(self):
+        swift.BaseApp("name", QObject())
+
+    def test_frames(self):
+        self.assertIsInstance(self.app.frames(), Iterable)
+
+    def test_broadcast(self):
+        self.app.broadcast("ch1", "msg")
+        self.app.broadcast("ch1", lambda: None)
```
kangz12345, line 159: What is this code for?

tests/test_swift.py
```diff
@@ -87,35 +73,139 @@ def setUp(self):
...
+
+    def test_received_message(self):
+        self.app.receivedSlot = MagicMock()
+        self.app._receivedMessage("ch1", '"msg"')
+        self.app.receivedSlot.assert_called_once()
+        self.app._receivedMessage("ch1", '"msg1" "msg2"')
```
kangz12345, line 165: What is this code for?

tests/test_swift.py
```diff
@@ -87,35 +73,139 @@ def setUp(self):
...
+
+    def test_received_swiftcall_result(self):
+        self.app.swiftcall.update_result = MagicMock()
+        self.app._receivedSwiftcallResult(
+            "request", '{"done": true, "success": true, "value": null, "error": null}'
+        )
+        self.app.swiftcall.update_result.assert_called_once()
+        self.app._receivedSwiftcallResult(
+            "request", '{"done": "tr" "ue", "success": true, "value": null, "error": null}'
```
kangz12345, line 174: What is this code for?

tests/test_swift.py
BECATRUE: Umm.. I think it would be better not to use methods or classes for tests.
`swift.subscriberNames` is also a test subject, so I didn't use it.

tests/test_swift.py
BECATRUE: Oh, it isn't for the coverage.
I just checked if the types of the fields are correct.

If you think it isn't necessary, I will remove it.

tests/test_swift.py
BECATRUE: It is for checking to raise a `TypeError`.

tests/test_swift.py
BECATRUE: Similar to the above question, it is for checking to raise a `JSONDecodeError`.

tests/test_swift.py
BECATRUE: It is the same as the above!

tests/test_swift.py
```diff
@@ -87,35 +73,139 @@ def setUp(self):
...
+        )
+
+
+class SwiftcallProxyTest(unittest.TestCase):
```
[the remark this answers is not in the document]
BECATRUE: Okay! what do you think is lacking?

tests/test_swift.py
```diff
@@ -87,35 +73,139 @@ def setUp(self):
...
+            self.assertIn(name, appNames)
+
     def test_destroy_app(self):
         for name in APP_INFOS:
             self.swift.destroyApp(name)
             self.assertNotIn(name, self.swift._apps)
 
+    def test_update_frames(self):
+        self.swift._apps["app1"].frames.return_value = (QWidget(),)
+        self.swift.updateFrames("app1")
```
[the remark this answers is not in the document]
BECATRUE: Line 96 means that `frames()` of the app is updated because `QWidget()` is different from the original frame.
So I checked if `updateFrames()` runs correctly.

Howerver, as you said, the test cases seem to be unsufficient. I will update it.

tests/test_swift.py
```diff
@@ -87,35 +73,139 @@ def setUp(self):
...
         for name in APP_INFOS:
+            self.assertIn(name, self.swift._dockWidgets)
             self.assertIn(name, self.swift._apps)
         for channel in self.channels:
             self.assertIn(channel, self.swift._subscribers)
 
+    def test_app_names(self):
+        appNames = self.swift.appNames()
+        for name in APP_INFOS:
+            self.assertIn(name, appNames)
```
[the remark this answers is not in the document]
BECATRUE: Oh, you're right! I will implement more concretely.

tests/test_swift.py
BECATRUE: As I wrote in PR, the test for `subscribe()` wasn't implemented because it is called when the `swift` instance is initialized and I checked if `self.swift._subscribers` is correct.

Do you think it is not enough?

tests/test_swift.py
BECATRUE: Haha.. I thought it was the simplest way..

tests/test_swift.py
kangz12345: However, since the unit test should be independent, I think it should have its own unit test!
Maybe for the `createApp()` and `destroyApp()` as well.

tests/test_swift.py
kangz12345: Oh, it makes more sense! Let's stick to this code.

tests/test_swift.py
kangz12345: I think it is not necessary, because it is not our responsibility to check their type, but `PyQt5`'s.
We just used the `pyqtSignal()` so it should be the type that `PyQt5` package specifies.

tests/test_swift.py
kangz12345: It such cases, can't we use [`assertRaises()`](https://docs.python.org/3.7/library/unittest.html#unittest.TestCase.assertRaises)?

tests/test_swift.py
kangz12345: However it does not _check_ whether the `JSONDecodeError` occurred.
Can we know if it worked without exceptions or not?

tests/test_swift.py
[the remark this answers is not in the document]
kangz12345: For example, to test exact behaviors, e.g., the signal is emitted with proper contents, ignoring the method names starting with `_`, etc.

tests/test_swift.py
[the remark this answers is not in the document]
BECATRUE: Wow! It was a too important review. 
`Swift.removeFrame()` didn't run correctly because it didn't remove the frame from `self.swift._dockwidgets`.

tests/test_swift.py
BECATRUE: Okay! I will implement it.

tests/test_swift.py
BECATRUE: Oh, however, `broadcast()` doesn't emit this exception, but catch it as follows.
```python
def broadcast(self, channelName: str, content: Any):
        """Broadcasts the content to the target channel.

        Args:
            channelName: Target channel name.
            content: Content to be broadcast. It should be able to be converted to JSON object.
        """
        try:
            msg = json.dumps(content)
        except TypeError as e:
            print(f"swift.app.broadcast(): {e!r}")
        else:
            self.broadcastRequested.emit(channelName, msg)
```

So, we can't detect it using `assertRaises()`.

tests/test_swift.py
BECATRUE: Okay! Then, how about `test_init()` of `SwiftTest`?

kangz12345: Oh yeah those things as well 😄
Let's not instance-check them!

tests/test_swift.py
kangz12345: Oh I see... using `print()` seems awful 😨 
Then how about checking if `self.broadcastRequsted` is emitted?

tests/test_swift.py
BECATRUE: I have a question. 
The `createApp()` calls `subscribe()`, so I think if we test `createApp()`, it is not necessary to test `subscribe()`.
If you still want to sustain independently and separate into two tests, I think in the test for `createApp()`, `subscribe()` should be mocked.
What do you think of it?

BECATRUE: In `createApp()`, `subscribe()` and `addFrame()` are called, but I think it is enough to test all at once.
Do you agree with me..?

tests/test_swift.py
BECATRUE: Oh, it looks a good idea! I will do so.

tests/test_swift.py
kangz12345: I think mocking is the better option.
Reasons are:
1. In the future, we might change the implementation of `createApp()` or it could be removed.
2. Testing just `createApp()` might not cover enough cases for `subscribe()` (in general), since it could assume a special case, where `subscribe()` supports more general cases (just for example).

tests/test_swift.py
kangz12345: Then maybe checking whether `receivedSlot()` is called would be the only way...

tests/test_swift.py
kangz12345: Please make it a _test_, i.e., we can notice the unexpected results.

tests/test_swift.py
BECATRUE: Okay! Then do you think we need to test `_parseArgs()`, `_handleSwiftcall()`, and `_swiftcall()` individually?

BECATRUE:
I updated the reviews! But it isn't implemented yet about `subscribe()`.
And after you approve this PR, I will move out the test file as we discussed in #144.

tests/test_swift.py
kangz12345: Oh, then I will do that after this PR!

tests/test_swift.py
```diff
@@ -69,44 +69,64 @@ def setUp(self):
...
-        for name in APP_INFOS:
-            self.assertIn(name, appNames)
+        appNamesSet = set(self.swift.appNames())
+        self.assertEqual(appNamesSet, set(APP_INFOS))
+
+    def test_create_app(self):
+        app_ = MagicMock()
+        app_.frames.return_value = (QWidget(),)
+        cls = MagicMock(return_value=app_)
+        setattr(importlib.import_module.return_value, "cls3", cls)
```
kangz12345, line 90: Can you check if the `import_module()` is called with the argument `"module3"`?

tests/test_swift.py
```diff
@@ -69,44 +69,64 @@ def setUp(self):
...
+        self.swift.createApp(
+            "app3",
+            swift.AppInfo(**{"module": "module3", "cls": "cls3", "channel": ["ch1"]})
+        )
+        self.assertIn("app3", self.swift._apps)
+        self.assertIn("app3", self.swift._dockWidgets)
+        self.assertIn("app3", self.swift._subscribers["ch1"])
```
kangz12345, line 97: Can you check if the `self.swift._apps["app3"]` is an instance of `cls` (i.e., "cls3")?

tests/test_swift.py
```diff
@@ -69,44 +69,64 @@ def setUp(self):
...
+                self.assertNotIn(name, self.swift._subscribers[channel])
 
     def test_update_frames(self):
-        self.swift._apps["app1"].frames.return_value = (QWidget(),)
+        orgFramesSet = {dockWidget.widget() for dockWidget in self.swift._dockWidgets["app1"]}
+        newFrames = (QWidget(),)
+        newFramesSet = set(newFrames)
+        self.swift._apps["app1"].frames.return_value = newFrames
         self.swift.updateFrames("app1")
+        curFramesSet = {dockWidget.widget() for dockWidget in self.swift._dockWidgets["app1"]}
```
kangz12345, line 113: How about `finalFramesSet`? The term "original" and "current" look similar.

tests/test_swift.py
```diff
@@ -69,44 +69,64 @@ def setUp(self):
...
+        for frame in orgFramesSet - newFramesSet:
+            self.assertNotIn(frame, curFramesSet)
+        for frame in newFrames:
+            self.assertIn(frame, curFramesSet)
```
kangz12345, line 117: Isn't it equivalent to use `assertEqual()` for `newFramesSet` and `curFramesSet`?

tests/test_swift.py
```diff
@@ -69,44 +69,64 @@ def setUp(self):
...
     def test_subscriber_names(self):
         for channel in self.channels:
-            subscriberNames = self.swift.subscriberNames(channel)
-            for name, info in APP_INFOS.items():
-                if channel in info.channel:
-                    self.assertIn(name, subscriberNames)
+            subscriberNamesSet = set(self.swift.subscriberNames(channel))
+            self.assertEqual(
+                subscriberNamesSet,
+                {name for name, info in APP_INFOS.items() if channel in info.channel}
```
kangz12345, line 128: Great use of set comprehension!

tests/test_swift.py
```diff
@@ -155,24 +170,41 @@ def test_frames(self):
         self.assertIsInstance(self.app.frames(), Iterable)
 
     def test_broadcast(self):
+        self.app.broadcastRequested = MagicMock()
         self.app.broadcast("ch1", "msg")
+        self.app.broadcastRequested.emit.assert_called_once_with("ch1", '"msg"')
+
+    def test_broadcast_exception(self):
```
kangz12345, line 177: Nice separation of test cases!

tests/test_swift.py
BECATRUE: Sure! It can be checked just using `import_module.assert_called_with("module3")`.

tests/test_swift.py
BECATRUE: I think it is the best way to set a name attribute of `app_` as "cls3" and check it.

tests/test_swift.py
BECATRUE: Oh, it looks more obvious!

tests/test_swift.py
BECATRUE: You are right! I checked it out in a complicated way..

BECATRUE: Umm.. I think we should test both cases where `newFrames` contains `orgFrames` (the current test) it does not.

tests/test_swift.py
```diff
@@ -104,17 +108,23 @@ def test_destroy_app(self):
...
                 self.assertNotIn(name, self.swift._subscribers[channel])
 
-    def test_update_frames(self):
+    def test_update_frames_inclusive(self):
+        orgFramesSet = {dockWidget.widget() for dockWidget in self.swift._dockWidgets["app1"]}
+        newFramesSet = orgFramesSet | {QWidget()}
+        self.swift._apps["app1"].frames.return_value = tuple(newFramesSet)
+        self.swift.updateFrames("app1")
+        finalFramesSet = {dockWidget.widget() for dockWidget in self.swift._dockWidgets["app1"]}
+        self.assertFalse(finalFramesSet & (orgFramesSet - newFramesSet))
```
kangz12345, line 117: By definition `orgFramesSet - newFramesSet` is empty.
I think this assertion should be done for general cases (not inclusive nor exclusive).

tests/test_swift.py
```diff
@@ -104,17 +108,23 @@ def test_destroy_app(self):
             for channel in info.channel:
                 self.assertNotIn(name, self.swift._subscribers[channel])
 
-    def test_update_frames(self):
+    def test_update_frames_inclusive(self):
```
kangz12345, line 111: I think the meaning of "inclusive" and "exclusive" here is not so obvious.
Writing a docstring for these tests would help the readers to understand.

kangz12345 (changes requested):

tests/test_swift.py
BECATRUE: Oh, you are right! I will remove the line.

tests/test_swift.py
BECATRUE: Okay! I will add some descriptions.

BECATRUE:
I applied your reviews!

kangz12345 (approved):
Great!

BECATRUE:
Oh, I move the test file out of tests directory!
Please re-check it. @kangz12345

kangz12345 (approved):
LGTM

PR #176, opened 2023-05-25
Resolve an import error in pip installation

BECATRUE:
To resolve the error, I implemented as follows:
- Make the `qiwis` directory and move `qiwis.py` in it. The structure is similar to `iquip` repo's.
- Modify the `qiwis` command to `qiwis.qiwis:main`.

Also, I made a pre-release for testing if it works well.
Please check the following command runs correctly.
```shell
$ pip install git+https://github.com/snu-quiqcl/qiwis.git@v2.0.2-alpha
$ qiwis  # probabily there is no setup.json, it will cause an error, but the command should be recognized well
$ python
>>> from qiwis import qiwis
>>> qiwis.BaseApp
<class 'qiwis.qiwis.BaseApp'>
```
The pre-release will be removed after this PR is completed.


kangz12345:
The pip installation and the commands work as well in my machine!
How does this solve the problem?? Should it be a package, not a flat module?
If so, let's make an `__init__.py` and `from qiwis import *` in it.

BECATRUE:
> How does this solve the problem?? Should it be a package, not a flat module?

Honestly, I don't know exactly why, but I guess that the files we want to distribute should be located in a package, and we should clarify the package name into `packages` of `setup.py`.

In [Writing the Setup Script](https://docs.python.org/3/distutils/setupscript.html#listing-whole-packages) description, it shows how to release a package, not only a module.

BECATRUE:
> If so, let's make an `__init__.py` and `from qiwis import *` in it.

Did you intend `from .qiwis import *`? I think it's a good idea!

In this case, however, `python -m qiwis.qiwis -s examples/setup.json` causes an warning due to an import procedure as follows.
> RuntimeWarning: 'qiwis.qiwis' found in sys.modules after import of package 'qiwis', but prior to execution of 'qiwis.qiwis'; this may result in unpredictable behaviour

Then, I suggest two solutions:
1. Use `python qiwis/qiwis.py -s examples/setup.json` instead of the original command.
2. Make `__main__.py` in `qiwis` package and write as below. Then, we can run it by `python -m qiwis -s examples/setup.json`.
```python
from .qiwis import main
main()
```

kangz12345:
> In [Writing the Setup Script](https://docs.python.org/3/distutils/setupscript.html#listing-whole-packages) description, it shows how to release a package, not only a module.

How about [this example](https://docs.python.org/3.7/distutils/introduction.html#a-simple-example)?

BECATRUE:
> > In [Writing the Setup Script](https://docs.python.org/3/distutils/setupscript.html#listing-whole-packages) description, it shows how to release a package, not only a module.
> 
> How about [this example](https://docs.python.org/3.7/distutils/introduction.html#a-simple-example)?

Oh, I will check the example!

kangz12345:
> Oh, I will check the example!

Oops I'm sorry it's a legacy module `distutils`.

kangz12345:
@BECATRUE
Please find out my solution here!
You can try:
1. uninstall your `qiwis` by
```
pip uninstall qiwis
```
2. install `qiwis` from `setup-test` tag by
```
pip install git+https://github.com/snu-quiqcl/qiwis.git@setup-test
```

Does this work for you as well?
cf) I use `python3 -m pip` instead of `pip`.

BECATRUE:
> @BECATRUE Please find out my solution here! You can try:
> 
> 1. uninstall your `qiwis` by
> 
> ```
> pip uninstall qiwis
> ```
> 
> 2. install `qiwis` from `setup-test` tag by
> 
> ```
> pip install git+https://github.com/snu-quiqcl/qiwis.git@setup-test
> ```
> 
> Does this work for you as well? cf) I use `python3 -m pip` instead of `pip`.

It works so well! As you think, I think releasing only the package looks neat.
Then, may I modify this like your test branch.

kangz12345:
> It works so well! As you think, I think releasing only the package looks neat. Then, may I modify this like your test branch?

Good! I only changed the `packages` to `py_modules` in `setup()`.

BECATRUE:
A lot of parts have changed, so I made a new PR.
Please come to that PR.

PR #177, opened 2023-05-28
Resolve an import error in pip installation using `py_modules`

### LINKED ISSUE #175: Resolve an import error in pip installation
I tried to resolve the Pylint action bug in [iquip](https://github.com/snu-quiqcl/iquip) project.
While looking for a solution, I found out the followings:
- Using `pip install -e` in the GitHub action may not work well.
- We have to use only `pip install`.
So we need to update the release using `pip install`, not `-e` option.

In the current status, however, `pip install ${url}` causes an error as follows:

Thus, I will resolve this error and make `pip install` available.

BECATRUE:
As @kangz12345 found out the solution in #176, I resolved the error using `py_modules` in `setup.py`.

I modified something as follows:
- The version to `2.0.2`
- The download url in accordance with the latest version
- How to use `qiwis`
- Remove not used comment in `numgen` example (It was used when we ran the `numgen` app directly)

This PR wil close #175.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ QIWIS (**Q**u**I**qcl **W**idget **I**ntegration **S**oftware) is a framework fo
 ## How to download and install
 One of the below:
 - Using released version (_recommended_)
-1. `pip install -e git+https://github.com/snu-quiqcl/qiwis.git@v2.0.1#egg=qiwis`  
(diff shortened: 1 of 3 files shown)


kangz12345 (approved):
LGTM!

Issue #182, opened 2023-06-05
Replace all the logging codes with `logging` API


kangz12345:
### Feature you want to implement
Replace all the logging codes, e.g., `print()` or sending messages to the logger app, with `logging` API codes, so the new logger app (see #181) can handle it.

### How the feature is implemented
Just identify the logging parts and replace them.

Issue #181, opened 2023-06-05
Logger app with `logging` module


kangz12345:
### Feature you want to implement
Change the logger app to show the logs generated by `logging`.

### How the feature is implemented
Nice reference: [`Logging Cookbook`](https://docs.python.org/3.7/howto/logging-cookbook.html#a-qt-gui-for-logging)

See #179 for more context.

Aijuh:
cannot assign myself in assignees menu

Aijuh:
Is logging library is included in logger.py by importing Baseapp from qiwis.py? Is it correct that the issue want logger.py get message from logging in qiwis.py? I am still reading reference and PyQt tutorials, so it might seem as weired question.

kangz12345:
> Is logging library is included in logger.py by importing Baseapp from qiwis.py?

If you want to use the `logging` module in the `logger` module (logger.py), you should import `logging` again.
That is, you should do the following in logger.py:
```python
import logging
```

> Is it correct that the issue want logger.py get message from logging in qiwis.py?

Yes. Currently the `LoggerApp` listens to the "log" channel for logging, but you will change that mechanism to handle the logs directly using the `logging` module.

I think you can try the following in order:
1. Make a simple handler (you can just use these: [useful handlers](https://docs.python.org/3.7/howto/logging.html#useful-handlers)) in logger.py or any other file (you can create a new file).
2. Check if it can catch the logs from `qiwis` module (qiwis.py).
3. Referring to the [reference](https://docs.python.org/3.7/howto/logging-cookbook.html#a-qt-gui-for-logging), make the logger app to print out the log messages.

If you want to just create a new module leaving the logger.py, feel free to.

PR #183, opened 2023-06-06
Log with `logging` module in `qiwis`

### LINKED ISSUE #180: `Qiwis` central logging using `logging`
Use `logging` module for the `Qiwis` system's central logging.

Replace all the print-out parts with `logging` API.
Then they will appear depending on the logging configuration.

### Additional context
See also: https://docs.python.org/3.7/howto/logging.html#configuring-logging-for-a-library

kangz12345:
I added logging parts.
One can check the logging as follows:

Screenshot 2023-06-06 at 12 24 01

(The stack trace is printed by the logger, and the program is not dead, i.e., the exception is caught normally.)

If you are not familiar to `logging`, then I can make a test script for you.

This closes #180.

### DIFF
--- a/qiwis.py
+++ b/qiwis.py
@@ -320,7 +320,10 @@ class Qiwis(QObject):
         try:
             subscribers.remove(app)
         except KeyError:
+            logger.error("The app %s tried to unsubscribe from %s, "
+                         "which it does not subscribe to", app, channel)
             return False
+        logger.info("The app %s unsubscribed from %s", app, channel)
         return True
 
     @pyqtSlot(str, str)
--- a/qiwis.py
+++ b/qiwis.py
@@ -360,6 +360,7 @@ class Qiwis(QObject):
         for name, arg in args.items():
             cls = signature.parameters[name].annotation
             parsedArgs[name] = loads(cls, arg) if issubclass(cls, Serializable) else arg
+        logger.debug("Parsed arguments %s to %s", args, parsedArgs)
         return parsedArgs
 
     def _handleQiwiscall(self, sender: str, msg: str) -> Any:
@@ -413,11 +414,14 @@ class Qiwis(QObject):
             value = self._handleQiwiscall(sender, msg)
         except Exception as error:  # pylint: disable=broad-exception-caught
             result = QiwiscallResult(done=True, success=False, error=repr(error))
+            logger.exception("Qiwiscall failed")
         else:
             if isinstance(value, Serializable):
                 value = dumps(value)
             result = QiwiscallResult(done=True, success=True, value=value)
+            logger.info("Qiwiscall success")
         self._apps[sender].qiwiscallReturned.emit(msg, dumps(result))
+        logger.info("Qiwiscall result is reported")
 
 
 class BaseApp(QObject):
--- a/qiwis.py
+++ b/qiwis.py
@@ -577,9 +577,11 @@ class QiwiscallProxy:  # pylint: disable=too-few-public-methods
             result = QiwiscallResult(done=False, success=False)
             msg = dumps(info)
             if msg in self.results:
-                logger.warning("Duplicate message is ignored: %s", msg)
+                logger.warning("Duplicate qiwiscall request: %s, "
+                               "the new result overwrites the previous one", msg)
             self.results[msg] = result
             self.requested.emit(msg)
+            logger.debug("Requested a qiwiscall: %s converted from %s", msg, info)
             return result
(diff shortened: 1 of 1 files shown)


qiwis.py
```diff
@@ -171,6 +178,7 @@ def load(self, appInfos: Mapping[str, AppInfo]):
         """
         for name, info in appInfos.items():
             self.createApp(name, info)
+        logger.info("Loaded %d app(s)", len(appInfos))
```
BECATRUE, line 181: Do you prefer '% operator' to 'f-string'?
Indeed, I like the latter😅

qiwis.py
```diff
@@ -266,6 +278,7 @@ def updateFrames(self, name: str):
             self.removeFrame(name, orgFrames[frame])
         for frame in newFramesSet - orgFramesSet:
             self.addFrame(name, frame, info)
+        logger.info("Updated frames: %d -> %d", len(orgFramesSet), len(newFramesSet))
```
BECATRUE, line 282: The information looks not enough.
How about listing the frames of both `orgFrameSet` and `newFrameSet`?

qiwis.py
```diff
@@ -287,7 +300,11 @@ def subscribe(self, app: str, channel: str):
             app: The name of the app which wants to subscribe to the channel.
             channel: The target channel name.
         """
-        self._subscribers[channel].add(app)
+        if app in self._subscribers[channel]:
```
BECATRUE, line 304: 👍

qiwis.py
```diff
@@ -303,7 +320,10 @@ def unsubscribe(self, app: str, channel: str) -> bool:
         try:
             subscribers.remove(app)
         except KeyError:
+            logger.error("The app %s tried to unsubscribe from %s, "
+                         "which it does not subscribe to", app, channel)
             return False
+        logger.info("The app %s unsubscribed from %s", app, channel)
```
BECATRUE, line 327: Why don't you write it down in the present tense?
In `subscribe()`, you did that!

BECATRUE (changes requested):
I left a few reviews!

qiwis.py
kangz12345: For other uses, I also prefer the 'f-string'!
However, the `logging` system does some [optimization](https://docs.python.org/3.7/howto/logging.html#optimization):
> Formatting of message arguments is deferred until it cannot be avoided.

If we use the f-string, it will be directly converted to a string, even if the message is not printed at all.

See also: [Logging Cookbook](https://docs.python.org/3.7/howto/logging-cookbook.html#using-particular-formatting-styles-throughout-your-application).
However, note that we cannot use other formatting options, e.g., `{}`-style, when we _construct_ each log message.

In short, we _have to_ use such `%`-style formatting if we want to exploit the optimization.

qiwis.py
kangz12345: Yes indeed.
I will include the class names of the frames.

qiwis.py
kangz12345: 😄 Nice observation, I think it is a matter of taste.
For me, the expression "subscribe to A" refers to a certain continued _status_, so since `subscribe()` is called, the app subscribes to the channel until it unsubscribes the channel.
However, "unsubscribing" is more like an event at a specific time point, so I used the past tense.

I think "subscribing" is also used as an event just like "unsubscribing", so it is totally fine to use the past tense for `subscribe()` for consistency!
Which one do you prefer?

qiwis.py
BECATRUE: Wow, I never thought it would be for this reason!
Now, never mind😃

qiwis.py
BECATRUE: I was convinced by the first explanation!
I think you can leave it as it is.

qiwis.py
kangz12345: Actually, I realized that `removeFrame()` and `addFrame()` will log the information instead.
So I updated `removeFrame()` to log the frame class name as well.

kangz12345: Screen Shot 2023-06-08 at 19 07 33

BECATRUE: Oh, you are right.
Then, the logging looks enough!

BECATRUE (approved):
LGTM👍

Then, please release this as the latest version!

kangz12345:
> Then, please release this as the latest version!

Actually, there are several issues (#181, #182 and #184) left.

PR #189, opened 2023-06-26
Global constant namespace

### LINKED ISSUE #185: Global constant namespace
To provide a global constant (read-only) namespace which can be accessed by any apps in a qiwis system.
For example, `artiq-proxy` server host ip and port information can be stored in it.
This will be initialized by the "setup.json" file.

When loading the set-up file, the global constant namespace is initialized by the contents in the file.
However, sharing a mutable object among the apps is against our philosophy.
Therefore, the namespace must be immutable, i.e., read-only.
[`namedtuple`](https://docs.python.org/3.7/library/collections.html#collections.namedtuple) seems a nice candidate for this..

### Additional context
Suggested by: @BECATRUE

kangz12345:
This closes #185.

As discussed in #185, I used `namedtuple` for the namespace.
Moreover, to make it _constant_, or _immutable_, I converted all the `list`s and `dict`s to `tuple`s and `MappingPrxoyType`s, respectively.

Since they become immutable, I decided to share a single namedtuple object across all the apps, by assigning it as an internal (protected) class attribute `_constants`.
For the access, `BaseApp` now provides a read-only _property_ `constants`.

Since the global constant namespace is independent to `Qiwis` object, the setter function `set_global_constant_namespace()` is defined in module-level.

For the naming convention of `namedtuple`s, I mimicked the Python built-in type name styles, e.g., `contextmanager`, `unittest`, `namedtuple`, `defaultdict`, etc.

Now the setup.json can contain "constant" section as well.


qiwis.py
```diff
@@ -22,17 +22,21 @@
 import logging
 import os
 import sys
-from collections import defaultdict
+from collections import defaultdict, namedtuple
 from contextlib import contextmanager
+from types import MappingProxyType
```
BECATRUE, line 27: Wow.. It looks too strong tool!

qiwis.py
```diff
@@ -461,6 +467,11 @@ def __init__(self, name: str, parent: Optional[QObject] = None):
         self.received.connect(self._receivedMessage)
         self.qiwiscallReturned.connect(self._receivedQiwiscallResult)
 
+    @property
```
BECATRUE, line 470: 👍

qiwis.py
```diff
@@ -461,6 +467,11 @@ def __init__(self, name: str, parent: Optional[QObject] = None):
         self.received.connect(self._receivedMessage)
         self.qiwiscallReturned.connect(self._receivedQiwiscallResult)
 
+    @property
+    def constants(self) -> Tuple:
```
BECATRUE, line 471: Is it okay that the returning type is `Tuple`, not `NamedTuple`?

qiwis.py
```diff
@@ -609,6 +620,23 @@ def update_result(self, request: str, result: QiwiscallResult, discard: bool = T
         logger.debug("Qiwiscall result is updated: %s", _result)
 
 
+def set_global_constant_namespace(constants: Mapping[str, JsonType]) -> Tuple:
+    """Creates a immutable namedtuple and sets it as the global constant namespace.
```
BECATRUE: Please change 'a' to 'an'!😂

qiwis.py
```diff
@@ -609,6 +620,23 @@ def update_result(self, request: str, result: QiwiscallResult, discard: bool = T
...
+
+    Args:
+        constants: A mapping source for the global constant namespace.
+          The key-values become the constant name-values and hence the keys
+          should be a valid namedtuple field name.
+
+    Returns:
+        The created namedtuple, which is the global constant namespace.
+    """
+    constantnamespace = namedtuple("constantnamespace", constants.keys())
```
BECATRUE: How about starting the variable name with a capital letter?
In the [reference](https://docs.python.org/3/library/collections.html#collections.namedtuple), it looks common.

BECATRUE (changes requested):
I left a few reviews!

qiwis.py
kangz12345: I also wonder what the type hint should be, but [`typing`](https://docs.python.org/3/library/typing.html#other-special-directives) module documentation says:

> These functions and classes should not be used directly as annotations. Their intended purpose is to be building blocks for creating and declaring types.

Moreover, the namedtuples are `tuple` (you can confirm that with `isinstance()`).

qiwis.py
kangz12345: Oops, thank you!

qiwis.py
kangz12345: Oh, it makes sense! However, then we should change the pylint variable naming convention anyway.
Well, defining a class like that is quite okay in Python, so it might not be a problem.
If you agree, I will update the pylintrc files.

BECATRUE: Good! Please update the pylintrc files!

kangz12345:
It seems the GitHub action cannot install the dependencies well. I tried several times to pass.

qiwis.py
```diff
@@ -449,6 +453,8 @@ class BaseApp(QObject):
     qiwiscallRequested = pyqtSignal(str)
     qiwiscallReturned = pyqtSignal(str, str)
 
+    _constants = namedtuple("emptynamespace", ())
```
BECATRUE: Is the type of `_constants` correct?
I think you intened that `_constants` is a namedtuple object, but currently it is a namedtuple class!

How about changing it as follows?
```python
_constants = namedtuple("emptynamespace", ())()
```

BECATRUE: Also please add the description of `constants` in the docstring.

test.py
```diff
@@ -497,6 +505,67 @@ def test_dumps(self):
...
+        result = MappingProxyType({
+            "LIST": (None, 0, True, "list"),
+            "LIST_LIST": (0, (1, 2, (3, 4, 5))),
+            "DICT": MappingProxyType({"A": True, "B": False}),
+            "DICT_DICT": MappingProxyType({"A": 0, "B": MappingProxyType({"C": 1, "D": 2})}),
+            "LIST_DICT": (0, MappingProxyType({"A": 1, "B": (2, 3)})),
+        })
+        self.assertEqual(qiwis._immutable(source), result)
+        # when the root-type is list
+        self.assertEqual(qiwis._immutable(source["LIST_DICT"]), result["LIST_DICT"])
```
BECATRUE, line 567: 👍

BECATRUE, line 567: 👍

BECATRUE (changes requested):
This may be the final review!

qiwis.py
kangz12345: Nice review!

For the properties, I think we don't have to write the class docstring for it because we define them like methods.
We can write docstring in the property (method) definition.

Well, it would be good to make a rule "define properties before any other methods".

BECATRUE (approved):
LGTM!

PR #195, opened 2023-06-29
Logger app using `logging`

### LINKED ISSUE #181: Logger app with `logging` module

Aijuh:
This pull request is about #181

I set logger in logger.py as root logger, and logger in qiwis.py as __name__(__main__), so log from qiwis.py propagate to logger.py. Made loggingHandler to deal with log message and print to loggerFrame window. log message from qiwis.py printed in logger Frame.
Moreover I additionally add new Frame for selecting logging level (ex. DEBUG, ERROR .. ). If it is not needed, I will delete that part of code and pull request again.
Thank you.


examples/logger.py
```diff
@@ -3,13 +3,33 @@
...
 import time
+import logging
 from typing import Any, Optional, Tuple
 
 from PyQt5.QtCore import QObject, pyqtSlot, pyqtSignal
 from PyQt5.QtWidgets import QWidget, QVBoxLayout, QPushButton, QTextEdit, QLabel, QDialogButtonBox
 
 from qiwis import BaseApp
 
+logger = logging.getLogger()
```
kangz12345: Actually, we just want to add a handler to the root logger.
We don't have to define it as a global variable here.

examples/logger.py
```diff
@@ -3,13 +3,33 @@
...
+logger.setLevel(logging.INFO)
```
kangz12345: We should not change the root logger level. We only handle the log records. Changing the root logger might results in unexpected behaviors when using with other modules.

kangz12345: Therefore, please remove these two lines!

examples/logger.py
```diff
@@ -3,13 +3,33 @@
...
+
+class LoggingHandler(QObject, logging.Handler):
+    """
+    Handler for logger, handle input log by emit
+    """
```
kangz12345: We don't break lines for a single line docstring, for example:
`"""Handler for logger, handle input log by emit."""`

In addition, please finish your sentence with a period(.).

kangz12345: However, since we write a docstring for public signals, you should write it for `signal`.
For example:
```python
class LoggingHandler(...):
    """Handler for logger ...

    Signals:
        signal(log): A formatted log message is emitted.
    """
```

You can check other example docstrings in qiwis.py as well (e.g., `Qiwis`).

examples/logger.py
```diff
@@ -3,13 +3,33 @@
...
+    signal = pyqtSignal(str)
```
kangz12345, line 21: Please make a single empty line after the class docstring.

examples/logger.py
```diff
@@ -3,13 +3,33 @@
...
+    def __init__(self, slotfunc, *args, **kwargs):
```
kangz12345: Please make a single empty line between class-level definitions (after `signal = pyqtSignal(str)`).

examples/logger.py
```diff
@@ -3,13 +3,33 @@
...
 
 from PyQt5.QtCore import QObject, pyqtSlot, pyqtSignal
 from PyQt5.QtWidgets import QWidget, QVBoxLayout, QPushButton, QTextEdit, QLabel, QDialogButtonBox
 
 from qiwis import BaseApp
 
+logger = logging.getLogger()
+logger.setLevel(logging.INFO)
+
+class LoggingHandler(QObject, logging.Handler):
```
kangz12345: Of course you might want to make the code simpler by merging the `Signaller` in the example with the handler.
However, multiple inheritance could bring us subtle problems such as confusing method name resolution, and `super()` calls.
I would recommend to separate the `QObject` class and `logging.Handler` class, just like the example.

examples/logger.py
```diff
@@ -3,13 +3,33 @@
...
+    """
+    Handler for logger, handle input log by emit
+    """
+    signal = pyqtSignal(str)
+    def __init__(self, slotfunc, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
kangz12345: From this code, the reader is confused whether it calls the `__init__` of `QObject` or `logging.Handler`. If you inherit only one class, then the problem is resolved.

examples/logger.py
```diff
@@ -3,13 +3,33 @@
...
+        self.signal.connect(slotfunc)
+
+    def emit(self, record):
+        """ 
+        emit input signal to connected function
+        """
```
kangz12345: 1. Do not break the line after the first `"""`.
2. Use a capital letter for the first character.
3. Use the 3rd person singular verbs (emits, not emit) for function/method docstrings. It is easy to think the subject of the sentence is "it" and you omit it.
4. Finish sentences with a period.
5. Write the docstring for arguments.

Following the [style guide](https://google.github.io/styleguide/pyguide.html), the docstring should be:
```python
        """Emits input signal to connected function.
        
        Args:
            record: The emitted log record.
        """
```

However, since this is _overriding_ the method of `logging.Handler`, we can write as (see also: #44):
```python
        """Overridden."""
```

examples/logger.py
```diff
@@ -64,7 +86,48 @@ def buttonCancelClicked(self):
         """Clicks Cancel not to clear log."""
         self.close()
 
+class SetLevelFrame(QWidget):
+    """
+    Extended.
+    """
+    def __init__(self, parent: Optional[QObject] = None):
+
```
kangz12345: You should write the `"""Extended."""` docstring in the `__init__()`.

examples/logger.py
```diff
@@ -85,11 +148,21 @@ def __init__(self, name: str, parent: Optional[QObject] = None):
...
         self.confirmFrame.confirmed.connect(self.clearLog)
+        self.loggerFrame.levelButton.clicked.connect(self.selectLevel)
+        self.setlevelFrame = SetLevelFrame()
+        # define handler
+        self.handler = LoggingHandler(self.addLog)
+        # arbitrary format
+        fs ="%(name)s %(message)s"
+        formatter = logging.Formatter(fs)
+        self.handler.setFormatter(formatter)
+        logger.addHandler(self.handler)
```
kangz12345, line 135: Note that you only need the root logger object here.
Therefore, you can just write:
```python
logging.getLogger().addHandler(self.handler)
```

qiwis.py
```diff
@@ -37,7 +37,6 @@
 
 logger = logging.getLogger(__name__)
 
-
```
kangz12345: Please discard this change since it follows the rule (2 empty lines between global definitions).

examples/logger.py
```diff
@@ -115,6 +188,11 @@ def receivedSlot(self, channelName: str, content: Any):
         else:
             print(f"The message was ignored because "
                   f"the treatment for the channel {channelName} is not implemented.")
+            
+    @pyqtSlot()
+    def selectLevel(self):
+        """"shows a level select frame"""
```
kangz12345: 1. Start with a capital letter.
2. Finish with a period.

kangz12345 (changes requested):
Great job implementing the log level selection frame as well, but it is not a part of this PR and seems not that simple feature (there are some more design issues to decide). Please remove that feature in this PR.

In addition, please use [type annotations](https://peps.python.org/pep-0484/) in function or method definitions.

Aijuh:
Thank you. 
I will follow your instruction and fix the code.

Aijuh:
I changed my code and update it.
Should I make another pull request or just use this pull request?

kangz12345:
You can stay in this PR. The reviewers review and you reply to that and apply the reviews, and we repeat this process. At the end, when the reviewers approve this PR, then you can merge it.

examples/logger.py
```diff
@@ -3,13 +3,38 @@
...
+import logging
 from typing import Any, Optional, Tuple
 
 from PyQt5.QtCore import QObject, pyqtSlot, pyqtSignal
 from PyQt5.QtWidgets import QWidget, QVBoxLayout, QPushButton, QTextEdit, QLabel, QDialogButtonBox
 
 from qiwis import BaseApp
 
+class Signaller(QObject):
+    """Signal for LoggingHandler"""
```
kangz12345: Period at the end!

examples/logger.py
```diff
@@ -3,13 +3,38 @@
...
+    signal = pyqtSignal(str)
+
+class LoggingHandler(QObject, logging.Handler):
+    """Handler for logger.
+
+    Sends log message to connected function using emit
```
BECATRUE: It's just an option, but please add 'a' or 'the' properly.

BECATRUE: Also, please add the period at the end.

examples/logger.py
```diff
@@ -29,7 +54,6 @@ def __init__(self, parent: Optional[QObject] = None):
         layout.addWidget(self.logEdit)
         layout.addWidget(self.clearButton)
 
-
```
BECATRUE: Please add _two_ empty lines between outermost classes and functions!
For example, refer to other [files](https://github.com/snu-quiqcl/qiwis/blob/develop/qiwis.py).

examples/logger.py
```diff
@@ -85,11 +108,19 @@ def __init__(self, name: str, parent: Optional[QObject] = None):
         self.loggerFrame.clearButton.clicked.connect(self.checkToClear)
         self.confirmFrame = ConfirmClearingFrame()
         self.confirmFrame.confirmed.connect(self.clearLog)
+        # define handler
+        self.handler = LoggingHandler(self.addLog)
+        # arbitrary format
```
BECATRUE: If you plan to update this format, please use _TODO_ docstring referred to the [styleguide](https://google.github.io/styleguide/pyguide.html#312-todo-comments).

examples/logger.py
```diff
@@ -3,13 +3,38 @@
...
+
+    Sends log message to connected function using emit
+
+    Signal:
+          signal: Helps to send message to destination.
+    """
+    signal = pyqtSignal(str)
+
+    def __init__(self, slotfunc, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
BECATRUE: First of all, please add the constructor docstring about arguments.
Also are both `args` and `kwargs` necessary?

examples/logger.py
BECATRUE: I also agree with @kangz12345.

examples/logger.py
```diff
@@ -3,13 +3,38 @@
...
+    """Signal for LoggingHandler"""
+    signal = pyqtSignal(str)
+
+class LoggingHandler(QObject, logging.Handler):
+    """Handler for logger.
+
+    Sends log message to connected function using emit
+
+    Signal:
+          signal: Helps to send message to destination.
```
BECATRUE: Where is the signal used? I wonder what you intended.

examples/logger.py
```diff
@@ -85,11 +108,19 @@ def __init__(self, name: str, parent: Optional[QObject] = None):
         self.loggerFrame.clearButton.clicked.connect(self.checkToClear)
         self.confirmFrame = ConfirmClearingFrame()
         self.confirmFrame.confirmed.connect(self.clearLog)
+        # define handler
+        self.handler = LoggingHandler(self.addLog)
+        # arbitrary format
+        fs ="%(name)s %(message)s"
+        formatter = logging.Formatter(fs)
+        self.handler.setFormatter(formatter)
+        logging.getLogger().addHandler(self.handler)
```
BECATRUE: Is the logger in `qiwis.py` being called in well?

It seems not to print the logs generated in `qiwis`.
Also I think an appropriate name is necessary in `getLogger()`.

BECATRUE (changes requested):
Good start! I left several reviews.

I couldn't check if the logging is working well because it doesn't show the logs created in `qiwis`.
Let me know if I did anything wrong.

examples/logger.py
Aijuh: I missed it. Thank you.

examples/logger.py
Aijuh: Good review. I should delete it.

examples/logger.py
Aijuh: In order to print log message, the logger level should be set as logging.INFO in logger.py, because logger.info is used in qiwis.py to propagate log message.

From @kangz12345 I told to test root logger, so I just used root logger. But in order to manage logger level like INFO, DEBUG, it seems that name needed.

 I will set the name as 'parent' temporarily, and change logger level to show you message is well printed in next update.

examples/logger.py
Aijuh: No, I just copy and paste example code and didn't check it. Thank you.

Aijuh:
Thank you for the detailed review. I fixed the code.

examples/logger.py
BECATRUE: Oh, you're right! I didn't know `logging.getLogger()` would return the root logger.
Thanks for your detailed description😃

examples/logger.py
BECATRUE: @Aijuh Please modify this part!

BECATRUE (changes requested):
Please check the unresolved reviews written by Jiyong.

Aijuh:
I understand that we should not change root logger's level to prevent error. 
So I set logger name as 'parent'(name can be changed depend on notation rule) and upload  the code.
`
        logger = logging.getLogger("parent")
        logger.setLevel(logging.INFO)
        logger.addHandler(self.handler)
`
logger level is temporarily set as logging.INFO to print logger.info in qiwis.

BECATRUE:
Can you explain what you planned about a logger name and logging level?
I think we should decide them first.

In my opinion, I think we can just define a logger in each module such as `qiwis.py` and each apps as follows.
```python
logger = logging.getLogger(__name__)
```
It is the recommended method in the [docs](https://docs.python.org/3/howto/logging.html#advanced-logging-tutorial).

In terms of a logging level, I think each module is responsible for its logger and in `logger.py` it should just show the passed logs.
If we want to set the minimum logging level in the logger GUI, we can use `self.handler.setLevel()` referred to the [docs](https://docs.python.org/ko/3/library/logging.html#logging.Handler.setLevel).

I wonder what @kangz12345 intended and how @Aijuh are going to implement.

Aijuh:
I am planning to set logger in logger.py as "parent" and set other module's logger using `logging.getLogger("parent").getchild(__name__)` (might be changed).
 If I set logger name like this, logger from child propagate to parent. It works well between logger.py and qiwis.py.
 Of course I should test for other modules later.

For setting logger level, what I understand about @BECATRUE 's suggestion is each module is responsible for its logger and logger level. In this case what I am planning is as follows. 
1. Set logger level of parent logger in logger.py as DEBUG. Then, any logger message from other module will be printed.
2. Set logger level of child logger separately in that module. If we set logger level as ERROR, and input logger as logger.info, the message didn't propagate to parent logger in logger.py.

In case of changing handler's level using `self.handler.setLevel()`, it can block INFO level logs from DEBUG level logger. But it cannot allow ERROR level logger to get DEBUG level logs. 
In other words, it cannot lower logger's level. 
So I think that we may change logger's level directly or fix logger's level as DEBUG, and change handler's level.

I will wait exact name convention of logger from @kangz12345 and @BECATRUE .

BECATRUE:
Thanks for your detailed explanation👍

First, I strongly agree with your idea about setting the logger levels.
Also setting all loggers as subloggers of a specific logger, i.e. `parent` logger is the best solution in terms of management and danger of setting the level of the root looger.

However, I think it is a problem that `logger.py`, not `qiwis.py` has the authority to set up the `parent` logger.
Also it is a big restriction that all users using this package should know that all loggers must be subloggers of the `parent` logger.
So I suggests to move the authority to `qiwis.py` as follows.
```python
# qiwis.py
qiwisLogger = logging.getLogger("parent")
logger = qiwisLogger.getChild(__name__)  # Same as the current logger

# logger.py
import qiwis
...
logger = qiwis.qiwisLogger

# Other apps
import qiwis
...
logger = qiwis.qiwisLogger.getChild(__name__)
```
The `logger.py` is just an app, not a part of `qiwis` package, so we should remove any dependence on `qiwis.py`'s position.

Finally, about the name of the parent logger, I think 'qiwis' is more appropriate because `parent` looks a liitle ambiguous.

After listening to Jiyong's opinion, let's make a final decision.

kangz12345:
Thank you for the discussion.
I also talked with @Aijuh offline, and we decided to do one of the following:
### 1. Use the root logger and modify its level.
This might cause some unexpected behavior when any other module uses the root logger as well.
However, such a module also takes a risk that the root logger can be used by any other module.
Therefore, we just need to document that this modifies the root logger level.

### 2. Construct a logger hierarchy and make the apps follow the naming convention.
This does not modify the root logger, so it can avoid the problem of the option 1.
However, it cannot catch the logs emitted by the loggers which do not follow the naming convention (consider the loggers of any third-party modules which do not care of `qiwis`).
Moreover, the way to set the logger to be a child logger of `qiwis` logger is quite inconvenient.

In my opinion, unless there is a convenient way to set the logger hierarchy and catch the third-party apps' logs and there is no more problem with modifying the root logger, the option 1 looks better.

BECATRUE:
Thanks for suggesting options!

I don't think setting the child logger in option 2 is convenient because we can make a helper function in `qiwis.py`.
But the issue looks critical that we cannot track the logs generated by several third-party modules.

Thus I also agree with option 1.

Aijuh:
Getting discussion with @kangz12345 at yesterday lab meeting, what I understood is as follows.
We should consider the case where other App importing qiwis and also importing other module (for ex temp), and that temp module also using logging system. We want to manage this logger message from temp too.
In that case, using naming convention for parent and child is useless because temp will not know that convention.

So The solution is like this.

1. Using root logger in logger.py
2. Not using root logger. 

In 1. we should change root logger's level as DEBUG and manage logger.handler to block message. This solution is quite simple but dealing with root logger may cause problem. There is no serious problem in codes, but if other module also using root logger and we change root logger's level, it would cause problem. Also if we set root logger's level as DEBUG, all logger's default level is set as DEBUG.  

In 2. what I am planning is using naming convention of parent and child relationship in qiwis Apps. And at specific app who wants to read imported module's log we can manually read message and deliver to app's logger. logger.py's logger is the highest parent, and qiwis.py will be its child. And all apps logger is child of qiwis.py's logger. 

But I don't know detailed rules for importing qiwis.py in other module. If I got something wrong please correct it. Thank you.

Aijuh:
I understand there are two options.

1. Using root logger in logger.py
2. Not using root logger.
3. 
In 1. we should change root logger's level as DEBUG and manage logger.handler to block message. This solution is quite simple but dealing with root logger may cause problem. There is no serious problem in codes, but if other module also using root logger and we change root logger's level, it would cause problem. If we set root logger's level as DEBUG, all logger's default level is set as DEBUG, which is not a severe problem.

In 2. what I am planning is using naming convention of parent and child relationship in qiwis Apps. And at specific app who wants to read imported module's log we can manually read message and deliver to app's logger. logger.py's logger is the highest parent, and qiwis.py will be its child. And all apps logger is child of qiwis.py's logger.

As @kangz12345 told me that it may be okay to use root level logger and @BECATRUE said that it would be much simpler way, I fixed the code in that way. The root logger's level is now DEBUG. Please inform me if I missed something. Thank you.

BECATRUE:
Good! And please apply the above reviews about code format and `Signaller` first.
Also, our final PR must meet the GitHub actions i.e. Pylint and unittest.
To check the failed cause, please click the 'Details' button in the below GitHub actions.

Aijuh:
Sorry for the redundant code change. Dealing with pylint.error, I misunderstood location of blank space.
I changed commitment to follow code format and add comment about signaller.

kangz12345:
1. You can install `pylint` and test it with your own machine (you don't have to push and wait for the GitHub action to finish every time).
2. If you are done and want more feedback, you can re-request the review. If you don't have permission to request a review, please let me know.
3. Is it still not clear about the blank space rules, or now you understand it?

Aijuh:
> * You can install `pylint` and test it with your own machine (you don't have to push and wait for the GitHub action to finish every time).

I didn't know that, I will do in that way.


> * If you are done and want more feedback, you can re-request the review. If you don't have permission to request a review, please let me know.

I re-requested to reviewers.

> * Is it still not clear about the blank space rules, or now you understand it?

I understand it. I delete unnecessary blank space.

examples/logger.py
```diff
@@ -3,20 +3,51 @@
 """
 
 import time
+import logging
 from typing import Any, Optional, Tuple
-
 from PyQt5.QtCore import QObject, pyqtSlot, pyqtSignal
 from PyQt5.QtWidgets import QWidget, QVBoxLayout, QPushButton, QTextEdit, QLabel, QDialogButtonBox
-
```
kangz12345: You should keep these blank lines.
Please read [PEP8](https://peps.python.org/pep-0008/#imports).
Also, I improved the developer guide #2.

examples/logger.py
```diff
@@ -3,20 +3,51 @@
...
 from qiwis import BaseApp
 
+
+class Signaller(QObject):
+    """Signal for LoggingHandler."""
```
kangz12345: We put a single blank line after the class docstring.
I wrote this in #2.

Therefore:
```python
class Signaller(QObject):
    """Signal ..."""

    signal = pyqtSignal(str)


class LoggingHandler(...):
...
```

examples/logger.py
```diff
@@ -3,20 +3,51 @@
...
+    signal = pyqtSignal(str)
+
+
+class LoggingHandler(logging.Handler):
+    """Handler for logger.
+
+    Sends a log message to connected function using emit.
+
+    Signals:
+        signal(log): A formatted log message is emitted.
```
kangz12345: Please move this to the docstring of `Signaller`.

examples/logger.py
```diff
@@ -3,20 +3,51 @@
...
+    """Handler for logger.
+
+    Sends a log message to connected function using emit.
+
+    Signals:
+        signal(log): A formatted log message is emitted.    
+    """
+
+    def __init__(self, slotfunc):
+        """Connect an input function to the signal.
```
kangz12345: You should start this docstring with "Extended" as described in #44.
For example:
```python
"""Extended.

Connects an input function to the signal.

Args:
    slotfunc: A slot function connected to the signaller's signal.
"""
```

examples/logger.py
```diff
@@ -3,20 +3,51 @@
...
+
+        Args:
+            slotfunc: method function connected to signal.
+        """
+        super().__init__()
+        self.signaller = Signaller()
+        self.signaller.signal.connect(slotfunc)
+
+    def emit(self, record: logging.LogRecord):
+        """ Emits input signal to connected function."""
```
kangz12345: Please remove the heading whitespace (here -> """_Emits).

examples/logger.py
```diff
@@ -85,11 +117,21 @@ def __init__(self, name: str, parent: Optional[QObject] = None):
         self.loggerFrame.clearButton.clicked.connect(self.checkToClear)
         self.confirmFrame = ConfirmClearingFrame()
         self.confirmFrame.confirmed.connect(self.clearLog)
+        # define handler
```
kangz12345: You don't have to write obvious comments (the code immediately tells you that a handler is defined).

examples/logger.py
kangz12345: Please make this a TODO comment, for example `# TODO(aijuh): Change the log format when it is determined.`

examples/logger.py
```diff
@@ -85,11 +117,21 @@ def __init__(self, name: str, parent: Optional[QObject] = None):
...
+        self.handler = LoggingHandler(self.addLog)
+        # arbitrary format
+        fs ="%(name)s %(message)s"
+        formatter = logging.Formatter(fs)
+        self.handler.setFormatter(formatter)
+        logger = logging.getLogger()
+        logger.addHandler(self.handler)
+        logger.setLevel(logging.DEBUG)
```
kangz12345, line 134: You don't have to set the log level here. Let's set this later when you implement the level selector GUI.
It is okay that this cannot print out the info logs now.

qiwis.py
kangz12345: Please discard this change.

kangz12345 (changes requested):

Aijuh:
I will read reference ([PEP8](https://peps.python.org/pep-0008/#imports), #44, #2) to follow code format. Nice catch for what I missed. 

I changed for what I missed and fix the codes.

examples/logger.py
kangz12345: In addition, please notice that "Connects" and "A slot ...".

kangz12345:
Please check the unresolved comments.

examples/logger.py
```diff
@@ -3,20 +3,57 @@
...
+
+        Args:
+            slotfunc: method function connected to signal.
+        """
+        super().__init__()
+        self.signaller = Signaller()
+        self.signaller.signal.connect(slotfunc)
+
+    def emit(self, record: logging.LogRecord):
+        """Emits input signal to connected function."""
```
kangz12345: Since this _overrides_ the method `logging.Handler.emit()`, you should start it with `"""Overridden."""`.
For example:
```python
"""Overridden.

Emits ...
"""
```

kangz12345 (changes requested):

examples/logger.py
kangz12345: You should start a sentence with a capital letter.
Please notice the argument docstring of `slotfunc`.
I changed some expression.

FYI: We call the functions bounded by a class instances _methods_.

examples/logger.py
```diff
@@ -3,20 +3,57 @@
...
+    Sends a log message to connected function using emit.   
+    """
+
+    def __init__(self, slotfunc):
+        """Extended.
+
+        Connects the slotfunc to the signal.
+
+        Args:
+            slotfunc: method function connected to signal.
```
kangz12345: Please update this docstring as well.

kangz12345 (changes requested):

Aijuh:
Add overridden message because loggingHandler in logger.py override logging.Handler.emit

Delete duplicated comments

examples/logger.py
kangz12345: Please read this @Aijuh

Aijuh: I understand as adding argument docstring for `slotfunc`.

Since slotfunc for loggingHandler is addLog method in LoggerApp, I add that part.

examples/logger.py
```diff
@@ -3,20 +3,60 @@
...
+    Sends a log message to connected function using emit.   
+    """
+
+    def __init__(self, slotfunc):
+        """Extended.
+
+        Connects the slotfunc to the signal.
+
+        Args:
+            slotfunc: A method addLog in class LoggerApp.
```
kangz12345: Nice that you write it in a correct format, but the comment content is not appropriate.
The docstrings for public API (public classes, methods, functions, etc.) should be informative for users.
They should describe the general purpose and conditions for each function and its arguments.

In this case for `slotfunc`, your description is true but not informative for users.
Suppose that you are a user of this module. You want to use `LoggingHandler`. Would you think this docstring is helpful? It seems that it can only be used by `LoggerApp`, even if `LoggingHandler` is a public class.

From now on, please write the docstring as general as possible!

My suggestion: "A slot function for the log signal. It will be called with a log message when a log record is emitted."

examples/logger.py
```diff
@@ -3,20 +3,60 @@
...
+    signal = pyqtSignal(str)
+
+
+class LoggingHandler(logging.Handler):
+    """Handler for logger.
+
+    Sends a log message to connected function using emit.   
+    """
+
+    def __init__(self, slotfunc):
```
kangz12345: Please annotate type: `slotfunc: Callable[[str], None]`.

kangz12345: You have to add `Callable` in the `from typing import ...`.

kangz12345 (changes requested):
Might be the final review. Almost there!

examples/logger.py
Aijuh: I agree. Thanks for detailed instruction. My thought was short. I will follow exact format and think in general perspective.

I fixed codes

examples/logger.py
```diff
@@ -28,7 +28,7 @@ class LoggingHandler(logging.Handler):
     Sends a log message to connected function using emit.   
     """
 
-    def __init__(self, slotfunc):
+    def __init__(self, slotfunc: Callable[[str],None]):
```
kangz12345: Please put a space after a comma!
`Callable[[str], None]`

examples/logger.py
```diff
@@ -34,7 +34,7 @@ def __init__(self, slotfunc):
         Connects the slotfunc to the signal.
 
         Args:
-            slotfunc: A method addLog in class LoggerApp.
+            slotfunc: A slot function which is called when log record emitted.
```
kangz12345: 👍

kangz12345 (changes requested):

examples/logger.py
```diff
@@ -85,11 +126,19 @@ def __init__(self, name: str, parent: Optional[QObject] = None):
         self.loggerFrame.clearButton.clicked.connect(self.checkToClear)
         self.confirmFrame = ConfirmClearingFrame()
         self.confirmFrame.confirmed.connect(self.clearLog)
+        self.handler = LoggingHandler(self.addLog)
```
BECATRUE, line 128: Why did you make `handler` as an attribute of `LoggerApp`?
Is it necessary?

examples/logger.py
```diff
@@ -3,20 +3,60 @@
...
+import logging
+from typing import Any, Optional, Tuple, Callable
 
 from PyQt5.QtCore import QObject, pyqtSlot, pyqtSignal
 from PyQt5.QtWidgets import QWidget, QVBoxLayout, QPushButton, QTextEdit, QLabel, QDialogButtonBox
 
 from qiwis import BaseApp
 
+
+class Signaller(QObject):
```
BECATRUE: I still wonder why this class is necessary?
It has only one signal and I think you can move it into `LoggingHandler`.

examples/logger.py
```diff
@@ -3,20 +3,60 @@
...
+            slotfunc: A slot function which is called when log record emitted.
+        """
+        super().__init__()
+        self.signaller = Signaller()
+        self.signaller.signal.connect(slotfunc)
+
+    def emit(self, record: logging.LogRecord):
+        """Overridden.
+        
+        Emits input signal to connected function.
```
BECATRUE: It is a minor suggestion.
How about 'Emits the input signal to the connected function'.

I don't know well which is more appropriate..

examples/logger.py
BECATRUE: Is there any discussion about this with Jiyong?

BECATRUE (changes requested):
I left a few reviews!

examples/logger.py
Aijuh: Its because pyQt5.Qtcore.pyqtSignal object has no attribute 'connect'. So we need to set Qobject as parent of LoggingHandler to use signal.connect. So we need to set Logging.Handler and Qobject as parent of LoggingHandler, but @kangz12345 said that multiple inheritance may be misleading. 
He pointed out vague meaning for initialization function.

examples/logger.py
Aijuh: No as far as I check, it doesn't matter whether setting it as local or class variable.

Aijuh:
1. Fix error in format
2. In LoggerApp, handler and logger are not class variable.

Aijuh:
Unittest failed with following error.

_E: Failed to fetch http://azure.archive.ubuntu.com/ubuntu/pool/main/libi/libinput/libinput10_1.20.0-1ubuntu0.2_amd64.deb  404  Not Found [IP: 52.252.75.106 80]
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?___

It seems that fetch link has error. Can I do something to fix it?

BECATRUE:
> Unittest failed with following error.
> 
> _E: Failed to fetch http://azure.archive.ubuntu.com/ubuntu/pool/main/libi/libinput/libinput10_1.20.0-1ubuntu0.2_amd64.deb 404 Not Found [IP: 52.252.75.106 80] E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?___
> 
> It seems that fetch link has error. Can I do something to fix it?

Oh, we also realized this problem in other repos.
Can you add `sudo apt-get update` into both Pylint and unittest in 'Install dependencies'?
The [unittest](https://github.com/snu-quiqcl/iquip/blob/develop/.github/workflows/unittest.yml) in `iquip` will help you!

examples/logger.py
```diff
@@ -3,20 +3,60 @@
...
-from typing import Any, Optional, Tuple
+import logging
+from typing import Any, Optional, Tuple, Callable
 
 from PyQt5.QtCore import QObject, pyqtSlot, pyqtSignal
 from PyQt5.QtWidgets import QWidget, QVBoxLayout, QPushButton, QTextEdit, QLabel, QDialogButtonBox
 
 from qiwis import BaseApp
 
+
```
BECATRUE: Please remove the empty line.
We usually make only one line after imporing modules.

examples/logger.py
BECATRUE: Oh, I understand now!
Then, how about adding `_` at the begining of the class name? (Just rename `_Signaller`)
It is only for `LoggingHandler`, so it can be treated as a private class.

BECATRUE (commented):
it looks good!
After resolving the unittest problem, please re-request to Jiyong.

examples/logger.py
```diff
@@ -3,20 +3,59 @@
...
+    Sends a log message to connected function using emit.   
+    """
+
+    def __init__(self, slotfunc: Callable[[str], None]):
+        """Extended.
+
+        Connects the slotfunc to the signal.
+
+        Args:
+            slotfunc: A slot function which is called when log record emitted.
```
kangz12345: ... when _a_ log record _is_ emitted.

examples/logger.py
kangz12345: Maybe setting it as an attribute would be good because it allows users to access the handler object.

examples/logger.py
```diff
@@ -3,20 +3,59 @@
...
+    signal = pyqtSignal(str)
+
+
+class LoggingHandler(logging.Handler):
+    """Handler for logger.
+
+    Sends a log message to connected function using emit.   
+    """
+
+    def __init__(self, slotfunc: Callable[[str], None]):
```
kangz12345: Oh, I'm sorry to change, but `Callable[[str], Any]` would be better since we don't care about the return value of `slotfunc`.

kangz12345 (changes requested):

Aijuh:
Fixed code with reference to comments.

I will properly use 'a' or 'the'.

examples/logger.py
```diff
@@ -27,13 +27,13 @@ class LoggingHandler(logging.Handler):
...
 
-    def __init__(self, slotfunc: Callable[[str], None]):
+    def __init__(self, slotfunc: Callable[[str], Any]):
         """Extended.
 
         Connects the slotfunc to the signal.
 
         Args:
-            slotfunc: A slot function which is called when log record emitted.
+            slotfunc: A slot function which is called when a log record emitted.
```
kangz12345: Since the logger emits the log records, so "a log record _is_ emitted.", i.e., you should use the passive form using "is".

kangz12345 (changes requested):

.github/workflows/unittest.yml
```diff
@@ -16,6 +16,7 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Install dependencies
       run: |
+        sudo apt-get update
```
kangz12345, line 19: This causes a merge conflict. Please change it to `sudo apt-get clean`.
Don't change it for pylint.yml!

kangz12345 (changes requested):

Aijuh:
Fixed comments.

Sorry for the repeated grammer error. I will check every comment with the grammer program later.

Fix pylint and unittest.

kangz12345 (dismissed):
LGTM!

.github/workflows/unittest.yml
Aijuh: I think changing to `sudo apt-get clean` makes unittest error again.

kangz12345: Hmm, I think so too. 
Since there is no other PR in this repository, let's revert it to `sudo apt-get update` and resolve the merge conflict.
Please update your code, then I will resolve the conflict.

Aijuh: Updated code in unittest.

kangz12345: Good, I resolved the conflict.

BECATRUE (approved):
Great job adjusting to our coding style!

Then, please merge this branch to the develop branch just by clicking the below button😃

PR #201, opened 2023-07-13
Aijuh/199/log level selector

### LINKED ISSUE #199: Add log level selector
### Feature to implement

Add an additional widget to select log level.
 

### Implementation plan

1. Inside LoggerApp class, add a new combo widget(name :levelHandler) to select log level (DEBUG, INFO, WARNING, ERROR, CRITICAL)
2. Connect a new function(name : fixLevel) to combo widget inside the class LoggerApp
3. Inside the function fixLevel, modify Level of LoggerApp's handler.

### Additional Info

To modify logger level of LoggerApp using handler, logger's level should be set as 'DEBUG'. If logger's level is 'INFO', by changing level of handler, we can block 'INFO' or 'WARNING' signal. But 'DEBUG' signal cannot reach to logger by changing handler's level.
In other words handler can block more serious level, but cannot allow less serious level to pass.

Currently we are using root logger as LoggerApp's logger and change its level as 'DEBUG'. If we cannot change root logger's level, implementation should be changed.

Aijuh:
Finish implementing log level selector as @kangz12345 requested as a week goal.
Doubly checked comments and grammer issue. 
I will fix code right away if I missed something.

This closes #199

### DIFF
--- a/examples/logger.py
+++ b/examples/logger.py
@@ -7,7 +7,7 @@ import logging
 from typing import Any, Optional, Tuple, Callable
 
 from PyQt5.QtCore import QObject, pyqtSlot, pyqtSignal
-from PyQt5.QtWidgets import QWidget, QVBoxLayout, QPushButton, QTextEdit, QLabel, QDialogButtonBox
+from PyQt5.QtWidgets import QWidget, QVBoxLayout, QPushButton, QTextEdit, QLabel, QDialogButtonBox, QComboBox
 
 from qiwis import BaseApp
 
@@ -63,10 +63,12 @@ class LoggerFrame(QWidget):
         self.logEdit = QTextEdit(self)
         self.logEdit.setReadOnly(True)
         self.clearButton = QPushButton("Clear")
+        self.levelSelector = QComboBox(self)
         # layout
         layout = QVBoxLayout(self)
         layout.addWidget(self.logEdit)
         layout.addWidget(self.clearButton)
+        layout.addWidget(self.levelSelector)
 
 
 class ConfirmClearingFrame(QWidget):
@@ -133,6 +135,29 @@ class LoggerApp(BaseApp):
         logger = logging.getLogger()
         logger.setLevel(logging.DEBUG)
         logger.addHandler(self.handler)
+        self.initLevelSelector(self.loggerFrame.levelSelector)
+
+    def initLevelSelector(self, box: QComboBox):
+        """Initialize a combo box for selecting level."""
+        box.addItem("DEBUG")
+        box.addItem("INFO")
+        box.addItem("WARNING")
+        box.addItem("ERROR")
+        box.addItem("CRITICAL")
+        box.activated[str].connect(self.fixLevel)
+    
+    def fixLevel(self, text):
+        match text:
+            case "DEBUG":
+                self.handler.setLevel(logging.DEBUG)
+            case "INFO":
+                self.handler.setLevel(logging.INFO)        
+            case "WARNING":
+                self.handler.setLevel(logging.WARNING)
+            case "ERROR":
+                self.handler.setLevel(logging.ERROR)
+            case "CRITICAL":
+                self.handler.setLevel(logging.CRITICAL)
 
     def frames(self) -> Tuple[LoggerFrame]:
         """Overridden."""
--- a/examples/logger.py
+++ b/examples/logger.py
@@ -56,7 +56,7 @@ class LoggerFrame(QWidget):
     Attributes:
         logEdit: A textEdit which shows all logs.
         clearButton: A button for clearing all logs.
-        setLevelBox: A combo box for setting the logger's level.
+        setLevelBox: A comboBox for setting the logger's level.
     """
 
     def __init__(self, parent: Optional[QObject] = None):
@@ -142,7 +142,7 @@ class LoggerApp(BaseApp):
         self.loggerFrame.setLevelBox.textActivated.connect(self.setLevel)
 
     def setLevel(self, text: str):
-        """Responds to the levelSelector widget and changes the handler's level.
+        """Responds to the setLevelBox widget and changes the handler's level.
         
         Args:
             text: Selected level in the level select box.
@@ -150,9 +150,9 @@ class LoggerApp(BaseApp):
         level = {
             "DEBUG": logging.DEBUG,
             "INFO": logging.INFO,
-            "WARNING":logging.WARNING,
-            "ERROR":logging.ERROR,
-            "CRITICAL":logging.CRITICAL
+            "WARNING": logging.WARNING,
+            "ERROR": logging.ERROR,
+            "CRITICAL": logging.CRITICAL
         }
         if text in level:
             self.handler.setLevel(level[text])
--- a/examples/logger.py
+++ b/examples/logger.py
@@ -56,7 +56,7 @@ class LoggerFrame(QWidget):
     Attributes:
         logEdit: A textEdit which shows all logs.
         clearButton: A button for clearing all logs.
-        setLevelBox: A comboBox for setting the logger's level.
+        levelBox: A comboBox for setting the logger's level.
     """
 
     def __init__(self, parent: Optional[QObject] = None):
@@ -144,9 +144,11 @@ class LoggerApp(BaseApp):
     @pyqtSlot(str)
     def setLevel(self, text: str):
         """Responds to the setLevelBox widget and changes the handler's level.
-        
+
         Args:
             text: Selected level in the level select box.
+                  It should be one of 'DEBUG', 'INFO', 'WARNING', 'ERROR' and 'CRITICAL' (case-sensitive).
+                  Any other input is ignored.
         """
         level = {
             "DEBUG": logging.DEBUG,
--- a/examples/logger.py
+++ b/examples/logger.py
@@ -7,8 +7,9 @@ import logging
 from typing import Any, Optional, Tuple, Callable
 
 from PyQt5.QtCore import QObject, pyqtSlot, pyqtSignal
-from PyQt5.QtWidgets import QWidget, QVBoxLayout, QPushButton, QTextEdit, QLabel, QDialogButtonBox
-from PyQt5.QtWidgets import QComboBox
+from PyQt5.QtWidgets import (
+    QWidget, QVBoxLayout, QPushButton, QTextEdit, QLabel, QDialogButtonBox, QComboBox
+)
 
 from qiwis import BaseApp
 
@@ -55,6 +56,7 @@ class LoggerFrame(QWidget):
     Attributes:
         logEdit: A textEdit which shows all logs.
         clearButton: A button for clearing all logs.
+        setLevelBox: A combo box for setting the logger's level.
     """
 
     def __init__(self, parent: Optional[QObject] = None):
@@ -64,12 +66,12 @@ class LoggerFrame(QWidget):
         self.logEdit = QTextEdit(self)
         self.logEdit.setReadOnly(True)
         self.clearButton = QPushButton("Clear")
-        self.levelSelector = QComboBox(self)
+        self.setLevelBox = QComboBox(self)
         # layout
         layout = QVBoxLayout(self)
         layout.addWidget(self.logEdit)
         layout.addWidget(self.clearButton)
-        layout.addWidget(self.levelSelector)
+        layout.addWidget(self.setLevelBox)
 
 
 class ConfirmClearingFrame(QWidget):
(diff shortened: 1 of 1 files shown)


examples/logger.py
```diff
@@ -8,6 +8,7 @@
 
 from PyQt5.QtCore import QObject, pyqtSlot, pyqtSignal
 from PyQt5.QtWidgets import QWidget, QVBoxLayout, QPushButton, QTextEdit, QLabel, QDialogButtonBox
+from PyQt5.QtWidgets import QComboBox
```
kangz12345: If the import line becomes too long, you can use parenthesis:
```python
from module import (
    Class1, Class2, Class3,
    Class4, Class5, Class6, Class7,
)
```

BECATRUE: Please merge this line into the above one!

If the line is too long, write it down as follows:
```python
from PyQt5.QtWidgets import (
    QWidget, QVBoxLayout, ...
)
```

examples/logger.py
```diff
@@ -63,10 +64,12 @@ def __init__(self, parent: Optional[QObject] = None):
         self.logEdit = QTextEdit(self)
         self.logEdit.setReadOnly(True)
         self.clearButton = QPushButton("Clear")
+        self.levelSelector = QComboBox(self)
```
BECATRUE: Please add this widget into the attributes docstring!

kangz12345: We use the type of the widget as its variable name.
For example, `clearButton` is a push button, `logEdit` is a text edit.
Therefore, `levelBox` would be a valid candidate (you can use any other names that ends with box).

We follow this policy because it helps recognize the widget class instantaneously, and it describes the widget quite well.

examples/logger.py
```diff
@@ -133,6 +136,29 @@ def __init__(self, name: str, parent: Optional[QObject] = None):
         logger = logging.getLogger()
         logger.setLevel(logging.DEBUG)
         logger.addHandler(self.handler)
+        self.initLevelSelector(self.loggerFrame.levelSelector)
+
+    def initLevelSelector(self, box: QComboBox):
+        """Initializes a combo box for selecting level."""
```
BECATRUE: Please add the docstring about `box`.

kangz12345: You should write a argument docstring:

```python
"""Initializes ... level.

Args:
    box: ...
"""
```

examples/logger.py
```diff
@@ -133,6 +136,29 @@ def __init__(self, name: str, parent: Optional[QObject] = None):
...
+        box.addItem("DEBUG")
+        box.addItem("INFO")
+        box.addItem("WARNING")
+        box.addItem("ERROR")
+        box.addItem("CRITICAL")
+        box.activated[str].connect(self.fixLevel)
+
+    def fixLevel(self, text):
+        """Responds to the levelSelector widget and changes the handler's level."""
+        if text == "DEBUG":
```
BECATRUE: I suggest a simpler solution!

```python
self.handler.setLevel(getattr(logging, text))
```

examples/logger.py
```diff
@@ -133,6 +136,29 @@ def __init__(self, name: str, parent: Optional[QObject] = None):
...
         logger.addHandler(self.handler)
+        self.initLevelSelector(self.loggerFrame.levelSelector)
+
+    def initLevelSelector(self, box: QComboBox):
+        """Initializes a combo box for selecting level."""
+        box.addItem("DEBUG")
+        box.addItem("INFO")
+        box.addItem("WARNING")
+        box.addItem("ERROR")
+        box.addItem("CRITICAL")
```
kangz12345: You can use [`addItems()`](https://doc.qt.io/qt-5/qcombobox.html#addItems):
```python
box.addItems(("DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"))
```

examples/logger.py
```diff
@@ -133,6 +136,29 @@ def __init__(self, name: str, parent: Optional[QObject] = None):
...
+        box.activated[str].connect(self.fixLevel)
+
+    def fixLevel(self, text):
+        """Responds to the levelSelector widget and changes the handler's level."""
```
BECATRUE: Please add the docstring and type hint about `text`.

examples/logger.py
```diff
@@ -133,6 +136,29 @@ def __init__(self, name: str, parent: Optional[QObject] = None):
         logger = logging.getLogger()
         logger.setLevel(logging.DEBUG)
         logger.addHandler(self.handler)
+        self.initLevelSelector(self.loggerFrame.levelSelector)
+
+    def initLevelSelector(self, box: QComboBox):
+        """Initializes a combo box for selecting level."""
+        box.addItem("DEBUG")
```
BECATRUE: Referred to [addItems()](https://doc.qt.io/qtforpython-5/PySide2/QtWidgets/QComboBox.html#PySide2.QtWidgets.PySide2.QtWidgets.QComboBox.addItems), you can add all items in just a line!

examples/logger.py
```diff
@@ -133,6 +136,29 @@ def __init__(self, name: str, parent: Optional[QObject] = None):
...
+        box.addItem("INFO")
+        box.addItem("WARNING")
+        box.addItem("ERROR")
+        box.addItem("CRITICAL")
+        box.activated[str].connect(self.fixLevel)
```
kangz12345: Could you explain about the `activated[str]`? Is there any reference about this?
I found [`textActivated`](https://doc.qt.io/qt-5/qcombobox.html#textActivated).

examples/logger.py
```diff
@@ -133,6 +136,29 @@ def __init__(self, name: str, parent: Optional[QObject] = None):
...
+
+    def fixLevel(self, text):
```
kangz12345: Please annotate the type of `text`.

examples/logger.py
kangz12345: Please write the argument docstring.

examples/logger.py
BECATRUE: Referred to [activated](https://doc.qt.io/qtforpython-5/PySide2/QtWidgets/QComboBox.html#PySide2.QtWidgets.PySide2.QtWidgets.QComboBox.activated), this signal is deprecated.

Please use another function!
I suggest a solution.

```python
box.activated.connect(self.fixLevel)

def fixLevel(self):
    text = self.loggerFrame.levelSelector.currentText()
    ...
```

examples/logger.py
```diff
@@ -133,6 +136,29 @@ def __init__(self, name: str, parent: Optional[QObject] = None):
...
+        if text == "DEBUG":
+            self.handler.setLevel(logging.DEBUG)
+        if text == "INFO":
+            self.handler.setLevel(logging.INFO)
+        if text == "WARNING":
+            self.handler.setLevel(logging.WARNING)
+        if text == "ERROR":
+            self.handler.setLevel(logging.ERROR)
+        if text == "CRITICAL":
+            self.handler.setLevel(logging.CRITICAL)
```
kangz12345: 1. `text` can be any string and `fixLevel()` is a public method. In this case, you should notify that "nothing happens when an invalid `text` is given". Or, if you can guarantee that invalid inputs are never given, you can change it to an internal (private) method `_fixLevel()`.
2. Use `elif` for efficiency.

In this case, the code is simple enough so might not be a problem, but there are several other approaches for this kind of task (you may keep the current code):
- Use [`enum`](https://docs.python.org/3/library/enum.html)
- Use [`getattr()`](https://docs.python.org/3/library/functions.html#getattr)
- Use a dictionary:
```python
level = {
    "DEBUG": logging.DEBUG,
    "INFO": logging.INFO,
    ...
}[text]
```

examples/logger.py
BECATRUE: Also please move this line into the constructor!
We usually connect a signal to a slot in there.

examples/logger.py
BECATRUE: Oh, and please rename it to include 'Box`.
For legibilty, we always include the type in each widget's name.

BECATRUE (changes requested):
I left some reviews!

examples/logger.py
kangz12345: The name `fixLevel` can be interpreted as it fixes (freezes) the level as well.
How about `setLevel`?

kangz12345 (changes requested):

kangz12345:
Could you attach the screenshot of the GUI?

examples/logger.py
Aijuh: I am not sure whether signal 'activated' has problem, so I changed to 'currentTextChanged'. It emits signal whenever currentText is changed.
I moved codes to construtor.

examples/logger.py
Aijuh: I used this method at first, but after reading @kangz12345 's suggestion I used dictionary instead.

examples/logger.py
Aijuh: I used `textActivated` instead, not using `activated`.
I moved it to constructor.

Aijuh: I read [activated](https://doc.qt.io/qt-5/qcombobox.html#activated).
The signal 'activated[str]' emits signal as string when comboBox's current text is changed.  
I think textActivated is more appropriate. I will use it.

examples/logger.py
Aijuh: That seems reasonable. I will use it.

examples/logger.py
Aijuh: I will use dictionary and set function as public.

Aijuh:
Updated codes by reflecting feedback.
1. Removed the initialization function for setLevelBox and move its code to the class constructor.
2. Follow comment guideline and naming convention.
3. Change `activate[str]` to `textActivated`.
4. Change nested if-else structure to dictionary.

Following is screenshot of the frame
캡처

examples/logger.py
```diff
@@ -54,6 +56,7 @@ class LoggerFrame(QWidget):
     Attributes:
         logEdit: A textEdit which shows all logs.
         clearButton: A button for clearing all logs.
+        setLevelBox: A combo box for setting the logger's level.
```
BECATRUE: It's a minor suggestion, but how about changing 'combo box' into 'comboBox' similar to `logEdit`?

examples/logger.py
```diff
@@ -133,6 +138,24 @@ def __init__(self, name: str, parent: Optional[QObject] = None):
         logger = logging.getLogger()
         logger.setLevel(logging.DEBUG)
         logger.addHandler(self.handler)
+        self.loggerFrame.setLevelBox.addItems(["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"])
+        self.loggerFrame.setLevelBox.textActivated.connect(self.setLevel)
+
+    def setLevel(self, text: str):
+        """Responds to the levelSelector widget and changes the handler's level.
```
BECATRUE: Please change 'levelSelector' into 'setLevelBox'!

examples/logger.py
```diff
@@ -133,6 +138,24 @@ def __init__(self, name: str, parent: Optional[QObject] = None):
...
+        
+        Args:
+            text: Selected level in the level select box.
+        """
+        level = {
+            "DEBUG": logging.DEBUG,
+            "INFO": logging.INFO,
+            "WARNING":logging.WARNING,
+            "ERROR":logging.ERROR,
+            "CRITICAL":logging.CRITICAL
```
BECATRUE: Please add a space after `:`.

BECATRUE (changes requested):
Probably this may be a fianl review!

examples/logger.py
kangz12345: However, its argument is `int` (`index`). I wonder where you found the usage of the `[str]`.

examples/logger.py
```diff
@@ -63,10 +66,12 @@ def __init__(self, parent: Optional[QObject] = None):
         self.logEdit = QTextEdit(self)
         self.logEdit.setReadOnly(True)
         self.clearButton = QPushButton("Clear")
+        self.setLevelBox = QComboBox(self)
```
kangz12345: `setLevelBox` seems like a method name which sets the levelBox.
Of course it is for _setting_ the level, but it also _shows_ the current level.
I think `levelBox` describes its role well enough.

examples/logger.py
```diff
@@ -133,6 +138,24 @@ def __init__(self, name: str, parent: Optional[QObject] = None):
         logger = logging.getLogger()
         logger.setLevel(logging.DEBUG)
         logger.addHandler(self.handler)
+        self.loggerFrame.setLevelBox.addItems(["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"])
+        self.loggerFrame.setLevelBox.textActivated.connect(self.setLevel)
+
+    def setLevel(self, text: str):
```
kangz12345, line 145: How about adding the `pyqtSlot(str)` decorator?

kangz12345 (changes requested):

examples/logger.py
```diff
@@ -54,6 +56,7 @@ class LoggerFrame(QWidget):
     Attributes:
         logEdit: A textEdit which shows all logs.
         clearButton: A button for clearing all logs.
+        setLevelBox: A comboBox for setting the logger's level.
```
kangz12345: Update this to `levelBox`!

examples/logger.py
```diff
@@ -133,6 +138,25 @@ def __init__(self, name: str, parent: Optional[QObject] = None):
...
         logger.addHandler(self.handler)
+        self.loggerFrame.levelBox.addItems(["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"])
+        self.loggerFrame.levelBox.textActivated.connect(self.setLevel)
+
+    @pyqtSlot(str)
+    def setLevel(self, text: str):
+        """Responds to the setLevelBox widget and changes the handler's level.
+        
+        Args:
+            text: Selected level in the level select box.
```
kangz12345, line 149: Please add explanation about the input validity. For example:
`It should be one of DEBUG, INFO, WARNING, ERROR and CRITICAL (case-sensitive). Any other input is ignored.`

kangz12345 (changes requested):
Almost there!

examples/logger.py
Aijuh: Sorry. I should have double checked.

examples/logger.py
```diff
@@ -133,6 +138,27 @@ def __init__(self, name: str, parent: Optional[QObject] = None):
...
+        self.loggerFrame.levelBox.textActivated.connect(self.setLevel)
+
+    @pyqtSlot(str)
+    def setLevel(self, text: str):
+        """Responds to the setLevelBox widget and changes the handler's level.
+
+        Args:
+            text: Selected level in the level select box.
+                  It should be one of 'DEBUG', 'INFO', 'WARNING', 'ERROR' and 'CRITICAL'.
+                  it should be case-sensitive and any other input is ignored.
```
kangz12345: Please start with a capital letter: `It should be ...`.

examples/logger.py
```diff
@@ -133,6 +138,27 @@ def __init__(self, name: str, parent: Optional[QObject] = None):
...
+        self.loggerFrame.levelBox.addItems(["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"])
+        self.loggerFrame.levelBox.textActivated.connect(self.setLevel)
+
+    @pyqtSlot(str)
+    def setLevel(self, text: str):
+        """Responds to the setLevelBox widget and changes the handler's level.
+
+        Args:
+            text: Selected level in the level select box.
+                  It should be one of 'DEBUG', 'INFO', 'WARNING', 'ERROR' and 'CRITICAL'.
```
kangz12345: Not a big deal but please use double quotes `"` instead of `'`.

Aijuh:
Fixed mistakes.

kangz12345 (approved):

BECATRUE (approved):
LGTM

PR #206, opened 2023-07-17
Aijuh/182/log with logging

### LINKED ISSUE #182: Replace all the logging codes with `logging` API

Aijuh:
This closes #182.

I changed all codes using log channel into the codes using logging in poller.py, dbmgr.py, numger.py.
Format using 'f' is changed to '%' in logging to prevent pylint error. All logs' level is logging.info.
I think in cofig.json, channel name log should be deleted. But I leave it just in case.

### DIFF
--- a/examples/backend.py
+++ b/examples/backend.py
@@ -60,7 +60,7 @@ def read(db_path: str, table: str) -> Any:
                 f"SELECT * FROM {table} ORDER BY rowid DESC LIMIT 1"
             ).fetchone()[0]
     except sqlite3.Error as e:
-        logger.error("apps.backend.read(): %r", e)
+        logging.exception("apps.backend.read(): %r", e)
         return None
     finally:
         con.close()
@@ -100,7 +100,7 @@ def write(db_path: str, table: str, value: Any) -> bool:
                 (value,)
             )
     except sqlite3.Error as e:
-        logger.error("apps.backend.write(): %r", e)
+        logging.exception("apps.backend.write(): %r", e)
         return False
     finally:
         con.close()
--- a/examples/backend.py
+++ b/examples/backend.py
@@ -4,8 +4,12 @@ Backend module for offering various functions.
 
 import random
 import time
+import logging
 import sqlite3
 from typing import Any
+
+logger = logging.getLogger(__name__)
+
 
 def generate() -> int:
     """Generates a random number from 0 to 99.
@@ -56,7 +60,7 @@ def read(db_path: str, table: str) -> Any:
                 f"SELECT * FROM {table} ORDER BY rowid DESC LIMIT 1"
             ).fetchone()[0]
     except sqlite3.Error as e:
-        print(f"apps.backend.read(): {e!r}")
+        logger.error("apps.backend.read(): %r", e)
         return None
     finally:
         con.close()
@@ -96,7 +100,7 @@ def write(db_path: str, table: str, value: Any) -> bool:
                 (value,)
             )
     except sqlite3.Error as e:
-        print(f"apps.backend.write(): {e!r}")
+        logger.error("apps.backend.write(): %r", e)
         return False
     finally:
         con.close()
--- a/examples/backend.py
+++ b/examples/backend.py
@@ -100,7 +100,7 @@ def write(db_path: str, table: str, value: Any) -> bool:
                 (value,)
             )
     except sqlite3.Error:
-        logger.exception("Failed to write a value into table %s of database %s", db_path, table)
+        logger.exception("Failed to write a value into table %s of database %s", table, db_path)
         return False
     finally:
         con.close()
--- a/examples/backend.py
+++ b/examples/backend.py
@@ -100,7 +100,7 @@ def write(db_path: str, table: str, value: Any) -> bool:
                 (value,)
             )
     except sqlite3.Error:
-        logger.exception("Failed to write database %s from table %s.", db_path, table)
+        logger.exception("Failed to write a value into table %s of database %s", db_path, table)
         return False
     finally:
         con.close()
--- a/examples/backend.py
+++ b/examples/backend.py
@@ -59,8 +59,8 @@ def read(db_path: str, table: str) -> Any:
             value = con.execute(
                 f"SELECT * FROM {table} ORDER BY rowid DESC LIMIT 1"
             ).fetchone()[0]
-    except sqlite3.Error as e:
-        logging.exception("apps.backend.read(): %r", e)
+    except sqlite3.Error:
+        logger.exception("Failed to read table %s from database %s.", table, db_path)
         return None
     finally:
         con.close()
@@ -99,8 +99,8 @@ def write(db_path: str, table: str, value: Any) -> bool:
                 f"INSERT INTO {table} VALUES (?, datetime('now', 'localtime'))", 
                 (value,)
             )
-    except sqlite3.Error as e:
-        logging.exception("apps.backend.write(): %r", e)
+    except sqlite3.Error:
+        logger.exception("Failed to write database %s from table %s.", db_path, table)
         return False
     finally:
         con.close()
--- a/examples/datacalc.py
+++ b/examples/datacalc.py
@@ -141,11 +141,10 @@ class DataCalcApp(BaseApp):
         """
         dbBox = self.viewerFrame.dbBoxes[name]
         self.dbNames[name] = dbBox.currentText()
-        logger.info(
-            "Database %s is set as %s." % (name, self.dbNames[name]) 
-            if self.dbNames[name]
-            else "Database %s is not selected." %name
-        )
+        if self.dbNames[name]:
+            logger.info("Database %s is set as %s.", name, self.dbNames[name]) 
+        else: logger.info("Database %s is not selected.", name)
+        
 
     @pyqtSlot()
     def calculateSum(self):
--- a/examples/datacalc.py
+++ b/examples/datacalc.py
@@ -98,7 +98,7 @@ class DataCalcApp(BaseApp):
         newDBs = set([""])
         for db in content.get("db", ()):
             if any(key not in db for key in ("name", "path")):
-                logger.info("The message was ignored because "
+                logger.error("The message was ignored because "
                             "the database %s has no such key; name or path.", json.dumps(db))
                 continue
             name, path = db["name"], db["path"]
@@ -128,9 +128,9 @@ class DataCalcApp(BaseApp):
             if isinstance(content, dict):
                 self.updateDB(content)
             else:
-                logger.info("The message for the channel db should be a dictionary.")
+                logger.error("The message for the channel db should be a dictionary.")
         else:
-            logger.info("The message was ignored because "
+            logger.error("The message was ignored because "
                         "the treatment for the channel %s is not implemented.", channelName)
 
     @pyqtSlot(str)
@@ -145,7 +145,7 @@ class DataCalcApp(BaseApp):
         if self.dbNames[name]:
             logger.info("Database %s is set as %s.", name, self.dbNames[name])
         else:
-            logger.info("Database %s is not selected.", name)
+            logger.warning("Database %s is not selected.", name)
 
     @pyqtSlot()
     def calculateSum(self):
--- a/examples/datacalc.py
+++ b/examples/datacalc.py
@@ -141,10 +141,10 @@ class DataCalcApp(BaseApp):
         """
         dbBox = self.viewerFrame.dbBoxes[name]
         self.dbNames[name] = dbBox.currentText()
-        logging.info(
-            f"Database {name} is set as {self.dbNames[name]}."
+        logger.info(
+            "Database %s is set as %s." % (name, self.dbNames[name]) 
             if self.dbNames[name]
-            else f"Database {name} is not selected."
+            else "Database %s is not selected." %name
         )
 
     @pyqtSlot()
@@ -165,4 +165,4 @@ class DataCalcApp(BaseApp):
             result += value
         else:
             self.viewerFrame.numberLabel.setText(f"sum: {result}")
-            logger.log(f"Sum: {result}.")
+            logger.log("Sum: %f.", result)
--- a/examples/datacalc.py
+++ b/examples/datacalc.py
@@ -3,6 +3,7 @@ App module for showing the sum of two values from selected databases.
 """
 
 import os
+import json
 import logging
 import functools
 from typing import Any, Optional, Dict, Tuple
@@ -97,8 +98,8 @@ class DataCalcApp(BaseApp):
         newDBs = set([""])
         for db in content.get("db", ()):
             if any(key not in db for key in ("name", "path")):
-                print(f"The message was ignored because "
-                        f"the database {db} has no such key; name or path.")
+                logger.info("The message was ignored because "
+                        "the database %s has no such key; name or path.", json.dumps(db))
                 continue
             name, path = db["name"], db["path"]
             newDBs.add(name)
@@ -127,10 +128,10 @@ class DataCalcApp(BaseApp):
             if isinstance(content, dict):
                 self.updateDB(content)
             else:
-                print("The message for the channel db should be a dictionary.")
+                logger.info("The message for the channel db should be a dictionary.")
         else:
-            print(f"The message was ignored because "
-                  f"the treatment for the channel {channelName} is not implemented.")
+            logger.info("The message was ignored because "
+                        "the treatment for the channel %s is not implemented.", channelName)
 
     @pyqtSlot(str)
     def setDB(self, name: str):
@@ -141,9 +142,8 @@ class DataCalcApp(BaseApp):
         """
         dbBox = self.viewerFrame.dbBoxes[name]
         self.dbNames[name] = dbBox.currentText()
-        if self.dbNames[name]:
-            logger.info("Database %s is set as %s.", name, self.dbNames[name])
-        else: logger.info("Database %s is not selected.", name)
+        logger.info("Database %s is set as %s.", (name, self.dbNames[name]) if self.dbNames[name]
+                    else "Database %s is not selected.", name)
 
     @pyqtSlot()
     def calculateSum(self):
--- a/examples/datacalc.py
+++ b/examples/datacalc.py
@@ -99,7 +99,7 @@ class DataCalcApp(BaseApp):
         for db in content.get("db", ()):
             if any(key not in db for key in ("name", "path")):
                 logger.info("The message was ignored because "
-                        "the database %s has no such key; name or path.", json.dumps(db))
+                            "the database %s has no such key; name or path.", json.dumps(db))
                 continue
             name, path = db["name"], db["path"]
             newDBs.add(name)
@@ -142,8 +142,10 @@ class DataCalcApp(BaseApp):
         """
         dbBox = self.viewerFrame.dbBoxes[name]
         self.dbNames[name] = dbBox.currentText()
-        logger.info("Database %s is set as %s.", (name, self.dbNames[name]) if self.dbNames[name]
-                    else "Database %s is not selected.", name)
+        if self.dbNames[name]:
+            logger.info("Database %s is set as %s.", name, self.dbNames[name])
+        else:
+            logger.info("Database %s is not selected.", name)
 
     @pyqtSlot()
     def calculateSum(self):
@@ -163,4 +165,4 @@ class DataCalcApp(BaseApp):
             result += value
         else:
             self.viewerFrame.numberLabel.setText(f"sum: {result}")
-            logger.log("Sum: %f.", result)
+            logger.info("Sum: %f.", result)
--- a/examples/datacalc.py
+++ b/examples/datacalc.py
@@ -142,9 +142,8 @@ class DataCalcApp(BaseApp):
         dbBox = self.viewerFrame.dbBoxes[name]
         self.dbNames[name] = dbBox.currentText()
         if self.dbNames[name]:
-            logger.info("Database %s is set as %s.", name, self.dbNames[name]) 
+            logger.info("Database %s is set as %s.", name, self.dbNames[name])
         else: logger.info("Database %s is not selected.", name)
-        
 
     @pyqtSlot()
     def calculateSum(self):
--- a/examples/datacalc.py
+++ b/examples/datacalc.py
@@ -145,7 +145,7 @@ class DataCalcApp(BaseApp):
         if self.dbNames[name]:
             logger.info("Database %s is set as %s.", name, self.dbNames[name])
         else:
-            logger.warning("Database %s is not selected.", name)
+            logger.info("Database %s is not selected.", name)
 
     @pyqtSlot()
     def calculateSum(self):
--- a/examples/datacalc.py
+++ b/examples/datacalc.py
@@ -3,6 +3,7 @@ App module for showing the sum of two values from selected databases.
 """
 
 import os
+import logging
 import functools
 from typing import Any, Optional, Dict, Tuple
 
@@ -11,6 +12,9 @@ from PyQt5.QtWidgets import QWidget, QVBoxLayout, QComboBox, QPushButton, QLabel
 
 from qiwis import BaseApp
 from examples.backend import read
+
+logger = logging.getLogger(__name__)
+
 
 class ViewerFrame(QWidget):
     """Frame of for selecting databases and showing the calculated number.
@@ -137,8 +141,7 @@ class DataCalcApp(BaseApp):
         """
         dbBox = self.viewerFrame.dbBoxes[name]
         self.dbNames[name] = dbBox.currentText()
-        self.broadcast(
-            "log", 
+        logging.info(
             f"Database {name} is set as {self.dbNames[name]}."
             if self.dbNames[name]
             else f"Database {name} is not selected."
@@ -162,4 +165,4 @@ class DataCalcApp(BaseApp):
             result += value
         else:
             self.viewerFrame.numberLabel.setText(f"sum: {result}")
-            self.broadcast("log", f"Sum: {result}.")
+            logger.log(f"Sum: {result}.")
--- a/examples/datacalc.py
+++ b/examples/datacalc.py
@@ -99,7 +99,7 @@ class DataCalcApp(BaseApp):
         for db in content.get("db", ()):
             if any(key not in db for key in ("name", "path")):
                 logger.error("The message was ignored because "
-                            "the database %s has no such key; name or path.", json.dumps(db))
+                             "the database %s has no such key; name or path.", json.dumps(db))
                 continue
             name, path = db["name"], db["path"]
             newDBs.add(name)
@@ -131,7 +131,7 @@ class DataCalcApp(BaseApp):
                 logger.error("The message for the channel db should be a dictionary.")
         else:
             logger.error("The message was ignored because "
-                        "the treatment for the channel %s is not implemented.", channelName)
+                         "the treatment for the channel %s is not implemented.", channelName)
 
     @pyqtSlot(str)
     def setDB(self, name: str):
--- a/examples/dbmgr.py
+++ b/examples/dbmgr.py
@@ -3,6 +3,7 @@ App module for adding and removing available databases.
 """
 
 import os
+import logging
 from collections import namedtuple
 from typing import Optional, Tuple
 
@@ -11,6 +12,9 @@ from PyQt5.QtWidgets import (QWidget, QLabel, QPushButton, QFileDialog,
                              QHBoxLayout, QVBoxLayout, QListWidget, QListWidgetItem)
 
 from qiwis import AppInfo, BaseApp
+
+logger = logging.getLogger(__name__)
+
 
 class DBWidget(QWidget):
     """Widget for showing a database.
@@ -121,8 +125,7 @@ class DBMgrApp(BaseApp):
         """
         msg = {"db": [db._asdict() for db in self.dbList]}
         self.broadcast("db", msg)
-        self.broadcast(
-            "log",
+        logger.info(
             f"Database {name} is added." if isAdded
             else f"Database {name} is removed."
         )
--- a/examples/dbmgr.py
+++ b/examples/dbmgr.py
@@ -184,7 +184,7 @@ class DBMgrApp(BaseApp):
         """
         if self.openCloseDatacalcResult is not None:
             if not self.openCloseDatacalcResult.done:
-                print("DBMgrApp.openCloseDatacalc(): The previous qiwiscall must be done.")
+                logger.info("DBMgrApp.openCloseDatacalc(): The previous qiwiscall must be done.")
                 return
             if self.openCloseDatacalcResult.success:
                 self.isDatacalcOpen = not self.isDatacalcOpen
--- a/examples/dbmgr.py
+++ b/examples/dbmgr.py
@@ -125,10 +125,7 @@ class DBMgrApp(BaseApp):
         """
         msg = {"db": [db._asdict() for db in self.dbList]}
         self.broadcast("db", msg)
-        logger.info(
-            "Database %s is added." if isAdded
-            else "Database %s is removed.", name
-        )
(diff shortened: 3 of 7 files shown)


examples/datacalc.py
```diff
@@ -137,8 +141,7 @@ def setDB(self, name: str):
         """
         dbBox = self.viewerFrame.dbBoxes[name]
         self.dbNames[name] = dbBox.currentText()
-        self.broadcast(
-            "log", 
+        logging.info(
```
BECATRUE: Please use % formatting, too!

examples/datacalc.py
```diff
@@ -162,4 +165,4 @@ def calculateSum(self):
             result += value
         else:
             self.viewerFrame.numberLabel.setText(f"sum: {result}")
-            self.broadcast("log", f"Sum: {result}.")
+            logger.log(f"Sum: {result}.")
```
BECATRUE: The same as above.

examples/numgen.py
```diff
@@ -138,18 +142,17 @@ def receivedSlot(self, channelName: str, content: Any):
             else:
                 print("The message for the channel db should be a dictionary.")
         else:
-            print(f"The message was ignored because "
-                  f"the treatment for the channel {channelName} is not implemented.")
+            print("The message was ignored because "\
```
BECATRUE: Please remove the backslash.

BECATRUE (changes requested):
It looks good!

If possible, could you add additional logs in each app, e.g., the failure case to save number in `numgen.py`?
Of course, you can separate this into another issue!

Also, please remove the channel `log` from `config.json`.

Aijuh:
I updated codes. 
Deleted backslash, and changed % format.
As for error case message, all the number in logging, it has well defined range. So we cannot know error from its value. 
Otherwise I added logger messages for error cases if there was none.

BECATRUE:
Oh, after you finish to apply our reviews, please re-request to us by clicking the above reviewers tab!
It will make us check quickly.

examples/datacalc.py
```diff
@@ -137,12 +141,9 @@ def setDB(self, name: str):
...
         self.dbNames[name] = dbBox.currentText()
-        self.broadcast(
-            "log", 
-            f"Database {name} is set as {self.dbNames[name]}."
-            if self.dbNames[name]
-            else f"Database {name} is not selected."
-        )
+        if self.dbNames[name]:
+            logger.info("Database %s is set as %s.", name, self.dbNames[name])
+        else: logger.info("Database %s is not selected.", name)
```
BECATRUE: Please move the body to a new line!

BECATRUE (changes requested):
I think all `print()` statements should be replaced to use logging.
Indeed, we just printed all logs before because the logger is not completed yet😅

Aijuh:
I changed `print` to `logger`. But there is some issue.

I changed 'f' format including data type dictionary `db` to the '%' format as below by importing json.
There might be bad thing by importing json or I misused it.
`logger.info("The message was ignored because "`
`"the database %s has no such key; name or path.", json.dumps(db))`

Also in examples/backend.py, I couldn't figure out what is e!r in `print(f"apps.backend.write(): {e!r}")`, so I just leave it as before.
Can I just set it as `logger.info("apps.backend.write(): %s", {e!r})`?

BECATRUE:
> There might be bad thing by importing json or I misused it.
`logger.info("The message was ignored because "
"the database %s has no such key; name or path.", json.dumps(db))`

I think there is no problem because `json.dumps()` returns just a string!

BECATRUE:
> Can I just set it as `logger.info("apps.backend.write(): %s", {e!r})?`

I'm not sure, but I think it doesn't cause a problem.
I couldn't find the detailed documentation about `!r`, but it looks another way to use `repr()`.
In other words, `e|r` is the same as `repr(e)`.

For more details, please see this [documentation](https://docs.python.org/2/tutorial/inputoutput.html#fancier-output-formatting) and this [example](https://peps.python.org/pep-0654/#unmatched-exceptions).
I think it is just a convention to print an exception.

Aijuh:
Thank you for the reply.
I think the type format is string but sure about it. I will leave `print` just in case.

Aijuh:
Thank you for good reference. I changed {e!r} to %r.

examples/backend.py
```diff
@@ -56,7 +60,7 @@ def read(db_path: str, table: str) -> Any:
                 f"SELECT * FROM {table} ORDER BY rowid DESC LIMIT 1"
             ).fetchone()[0]
     except sqlite3.Error as e:
-        print(f"apps.backend.read(): {e!r}")
+        logger.error("apps.backend.read(): %r", e)
```
BECATRUE: 👍

examples/datacalc.py
```diff
@@ -93,8 +98,8 @@ def updateDB(self, content: dict):
         newDBs = set([""])
         for db in content.get("db", ()):
             if any(key not in db for key in ("name", "path")):
-                print(f"The message was ignored because "
-                        f"the database {db} has no such key; name or path.")
+                logger.info("The message was ignored because "
+                        "the database %s has no such key; name or path.", json.dumps(db))
```
BECATRUE: Please align the starting positions of the two lines of double quotation marks.

```python
logger.info("The message was ignored because "
            "the database %s has no such key; name or path.", json.dumps(db))
```

examples/datacalc.py
```diff
@@ -137,12 +142,8 @@ def setDB(self, name: str):
...
         dbBox = self.viewerFrame.dbBoxes[name]
         self.dbNames[name] = dbBox.currentText()
-        self.broadcast(
-            "log", 
-            f"Database {name} is set as {self.dbNames[name]}."
-            if self.dbNames[name]
-            else f"Database {name} is not selected."
-        )
+        logger.info("Database %s is set as %s.", (name, self.dbNames[name]) if self.dbNames[name]
+                    else "Database %s is not selected.", name)
```
BECATRUE: Does it run well..?
In my local, it seems to match the formatting incorrectly.

examples/logger.py
```diff
@@ -186,10 +186,10 @@ def receivedSlot(self, channelName: str, content: Any):
             if isinstance(content, str):
                 self.addLog(content)
             else:
-                print("The message for the channel log should be a string.")
+                self.logger.info("The message for the channel log should be a string.")
```
BECATRUE: Please use `logger = logging.getLogger(__name__)` not `self.logger`!

examples/numgen.py
```diff
@@ -109,8 +114,8 @@ def updateDB(self, content: dict):
         newDBs = set([""])
         for db in content.get("db", ()):
             if any(key not in db for key in ("name", "path")):
-                print(f"The message was ignored because "
-                        f"the database {db} has no such key; name or path.")
+                logger.info("The message was ignored because "
+                        "the database %s has no such key; name or path.", json.dumps(db))
```
BECATRUE: Indentation!

examples/numgen.py
```diff
@@ -136,20 +141,19 @@ def receivedSlot(self, channelName: str, content: Any):
             if isinstance(content, dict):
                 self.updateDB(content)
             else:
-                print("The message for the channel db should be a dictionary.")
+                logger.info("The message for the channel db should be a dictionary.")
         else:
-            print(f"The message was ignored because "
-                  f"the treatment for the channel {channelName} is not implemented.")
+            logger.info("The message was ignored because "
+                  "the treatment for the channel %s is not implemented.", channelName)
```
BECATRUE: Indentation!

examples/numgen.py
```diff
@@ -136,20 +141,19 @@ def receivedSlot(self, channelName: str, content: Any):
...
         """Sets the database to store the number."""
         self.dbName = self.generatorFrame.dbBox.currentText()
         self.viewerFrame.statusLabel.setText("database updated")
-        self.broadcast(
-            "log", 
-            f"Database to store is set as {self.dbName}." if self.dbName
-            else "Database to store is not selected."
+        logger.info(
+            "Database to store is set as %s." if self.dbName
+            else "Database to store is not selected.", self.dbName
```
BECATRUE: It seems to make a syntax error.

```python
logger.info(
            "Database to store is set as %s.", self.dbName if self.dbName
            else "Database to store is not selected."
)
```

Isn't this correct?

examples/poller.py
```diff
@@ -95,8 +100,8 @@ def updateDB(self, content: dict):
         newDBs = set([""])
         for db in content.get("db", ()):
             if any(key not in db for key in ("name", "path")):
-                print(f"The message was ignored because "
-                        f"the database {db} has no such key; name or path.")
+                logger.info("The message was ignored because "
+                        "the database %s has no such key; name or path.", json.dumps(db))
```
BECATRUE: Indentation!

examples/poller.py
```diff
@@ -122,26 +127,25 @@ def receivedSlot(self, channelName: str, content: Any):
...
     def setDB(self):
         """Sets the database to store the polled number."""
         self.dbName = self.viewerFrame.dbBox.currentText()
-        self.broadcast(
-            "log", 
-            f"Polled database is set as {self.dbName}." if self.dbName
-            else "Polled database is not selected."
+        logger.info(
+            "Polled database is set as %s." if self.dbName
+            else "Polled database is not selected.", self.dbName
```
BECATRUE: Please check the syntax error.

BECATRUE (changes requested):
Several codes are likely to make syntax errors.
It would be a tedious work, but please check if each statement doesn't cause an error.

examples/datacalc.py
kangz12345: For such cases, please use the `if-else` _statement_, not the _expression_ for readability (for all the other similar cases as well).
For example:
```python
if self.dbNames[name]:
    logger.info("Database %s is set as %s.", name, self.dbNames[name])
else:
    logger.info("Database %s is not selected.", name)
```

kangz12345:
For logging the exceptions, you can consider using [`exception()`](https://docs.python.org/3/library/logging.html#logging.Logger.exception).
This will (probably) print out the exception stack information as well.

examples/datacalc.py
```diff
@@ -162,4 +163,4 @@ def calculateSum(self):
             result += value
         else:
             self.viewerFrame.numberLabel.setText(f"sum: {result}")
-            self.broadcast("log", f"Sum: {result}.")
+            logger.log("Sum: %f.", result)
```
kangz12345: Why did you use `log()` instead of `info()` here?

examples/datacalc.py
Aijuh: I tested it and found log message printed well. But I will change the code as @kangz12345 suggested.

examples/logger.py
Aijuh: I remember I heard that we should not set logger in logger.py as global variable. But I will set it as global variable for now.

examples/numgen.py
Aijuh: I found 
_logger.info(
            "Database to store is set as %s." % self.dbName if self.dbName
            else "Database to store is not selected."
)_
print message as we want.

But pylint tell me to use 
_logger.info(
          f"Database to store is set as {self.dbName}." if self.dbName
            else "Database to store is not selected."
)_ 
again. So I used it.

examples/datacalc.py
Aijuh: My mistake. I will change to log.info for integrity.

Aijuh:
Fix codes.

examples/numgen.py
BECATRUE: Oh, then does Pylint say that my suggestion doesn't meet the Pylint condition?
I think not using the _f-formatting_ is better..

examples/datacalc.py
```diff
@@ -123,10 +128,10 @@ def receivedSlot(self, channelName: str, content: Any):
             if isinstance(content, dict):
                 self.updateDB(content)
             else:
-                print("The message for the channel db should be a dictionary.")
+                logger.info("The message for the channel db should be a dictionary.")
         else:
-            print(f"The message was ignored because "
-                  f"the treatment for the channel {channelName} is not implemented.")
+            logger.info("The message was ignored because "
+                        "the treatment for the channel %s is not implemented.", channelName)
```
BECATRUE: I think all loggings in this file should be set as `WARNING` or `ERROR`.

How about @kangz12345?

examples/dbmgr.py
```diff
@@ -181,7 +184,7 @@ def openCloseDatacalc(self):
         """
         if self.openCloseDatacalcResult is not None:
             if not self.openCloseDatacalcResult.done:
-                print("DBMgrApp.openCloseDatacalc(): The previous qiwiscall must be done.")
+                logger.info("DBMgrApp.openCloseDatacalc(): The previous qiwiscall must be done.")
```
BECATRUE: The same as the above.

examples/numgen.py
```diff
@@ -109,8 +114,8 @@ def updateDB(self, content: dict):
         newDBs = set([""])
         for db in content.get("db", ()):
             if any(key not in db for key in ("name", "path")):
-                print(f"The message was ignored because "
-                        f"the database {db} has no such key; name or path.")
+                logger.info("The message was ignored because "
+                            "the database %s has no such key; name or path.", json.dumps(db))
```
BECATRUE: The same as the above.

examples/numgen.py
```diff
@@ -161,12 +165,13 @@ def generateNumber(self):
...
         # save the generated number
         dbPath = self.dbs[self.dbName]
         is_save_success = write(os.path.join(dbPath, self.dbName), self.table, num)
         if is_save_success:
             self.viewerFrame.statusLabel.setText("number saved successfully")
-            self.broadcast("log", "Generated number saved.")
+            logger.info("Generated number saved.")
         else:
             self.viewerFrame.statusLabel.setText("failed to save number")
+            logger.info("Failed to save generated number")
```
BECATRUE: The same as the above.

examples/numgen.py
```diff
@@ -136,18 +141,17 @@ def receivedSlot(self, channelName: str, content: Any):
             if isinstance(content, dict):
                 self.updateDB(content)
             else:
-                print("The message for the channel db should be a dictionary.")
+                logger.info("The message for the channel db should be a dictionary.")
         else:
-            print(f"The message was ignored because "
-                  f"the treatment for the channel {channelName} is not implemented.")
+            logger.info("The message was ignored because "
+                        "the treatment for the channel %s is not implemented.", channelName)
```
BECATRUE: The same as the above.

examples/numgen.py
```diff
@@ -136,18 +141,17 @@ def receivedSlot(self, channelName: str, content: Any):
             if isinstance(content, dict):
                 self.updateDB(content)
             else:
-                print("The message for the channel db should be a dictionary.")
+                logger.info("The message for the channel db should be a dictionary.")
```
BECATRUE: The same as the above.

examples/logger.py
BECATRUE: Can you explain this more?

examples/poller.py
```diff
@@ -95,8 +100,8 @@ def updateDB(self, content: dict):
         newDBs = set([""])
         for db in content.get("db", ()):
             if any(key not in db for key in ("name", "path")):
-                print(f"The message was ignored because "
-                        f"the database {db} has no such key; name or path.")
+                logger.info("The message was ignored because "
```
BECATRUE: The level of all loggings of this file should be determined.

BECATRUE (changes requested):
I think we should talk about the level of all loggings, because in the most cases, we left a log for notifying a warning, not just for an information.

Also, please check the above two reviews I left before.

examples/numgen.py
kangz12345: As I said, please use `if-else` statement instead. I don't see any benefit from using the ternary expression.
In addition, `logger.info("%s" % x)` and `logger.info("%s", x)` are different! The former is just using the `%`-formatting, which should be replaced with the f-string as pylint complained. The latter is using the `logging` API. Comment questions if you need more explanation about this.

Aijuh:
I agree with that. I will wait @kangz12345's comment.

examples/numgen.py
kangz12345: In addition, previously the code used the ternary expression because it didn't use the `logging` API.

examples/datacalc.py
kangz12345: Oh, nice point! I wanted to point it out but I forgot.
@Aijuh please choose appropriate log levels and use `warning()`, `error()`, and `exception()` as well.

kangz12345 (changes requested):

examples/logger.py
Aijuh: I don't remember exact reason. But I heard that setting root logger as global variable is something to be avoided. 
I changed my previous plan of using global variable as below. 
1. In constructor of LoggerApp, set root logger as local variable and set its level as "DEBUG" and addHandler.
2. When we need logger of logger.py (i.e. root logger), just call root logger using logging.getLogger(). (line 189, 191)

kangz12345: You can use multiple loggers. The root logger is only for _reading_ the logs.
To _write_ logs in `loggerApp`, please use `logging.getLogger(__name__)`.

examples/numgen.py
Aijuh: @kangz12345 : I understand. I will use if-else instead.
@BECATRUE : When I print message as you suggested using `print` in python.idle, if `self.dbName == Sam`, it prints  _"Database to store is set as %s. Sam"_. But what we want is _"Database to store is set as Sam"_. I don't test for exact logger so I am not sure about case of logging.

Aijuh:
I refer to [logging](https://docs.python.org/3/howto/logging.html) 's log level.
I understand error as problem in a particular function, and warning as possible problem.
So  In the code below I choose warning because two case has same function for now.
```python
if self.dbNames[name]:
    logger.info("Database %s is set as %s.", name, self.dbNames[name])
else:
    logger.warning("Database %s is not selected.", name)

```

In the code below I use error because it has different function at that time.
```python
for db in content.get("db", ()):
    if any(key not in db for key in ("name", "path")):
        logger.error("The message was ignored because "
                     "the database %s has no such key; name or path.", json.dumps(db))
        continue
```

kangz12345:
> I refer to [logging](https://docs.python.org/3/howto/logging.html) 's log level. I understand error as problem in a particular function, and warning as possible problem. So In the code below I choose warning because two case has same function for now.
> 
> ```python
> if self.dbNames[name]:
>     logger.info("Database %s is set as %s.", name, self.dbNames[name])
> else:
>     logger.warning("Database %s is not selected.", name)
> ```

Actually in that case, I think `info()` is okay because it is a valid expected input (you can _unselect_ the database).
Of course, after all, databases should be determined so you can label it as a warning.
Conclusion: you can decide which to use. I just wanted to tell you that it is not an unexpected situation.

Aijuh:
Changed warning to info.

examples/logger.py
BECATRUE: I also agree with @kangz12345's suggestion!

examples/numgen.py
BECATRUE: Oh, thanks for notifying it.
But, if you adapt Jiyong's suggestion, we don't have to consider this problem more.

examples/datacalc.py
```diff
@@ -93,8 +98,8 @@ def updateDB(self, content: dict):
         newDBs = set([""])
         for db in content.get("db", ()):
             if any(key not in db for key in ("name", "path")):
-                print(f"The message was ignored because "
-                        f"the database {db} has no such key; name or path.")
+                logger.error("The message was ignored because "
+                            "the database %s has no such key; name or path.", json.dumps(db))
```
BECATRUE: Indentation!

examples/datacalc.py
```diff
@@ -123,10 +128,10 @@ def receivedSlot(self, channelName: str, content: Any):
             if isinstance(content, dict):
                 self.updateDB(content)
             else:
-                print("The message for the channel db should be a dictionary.")
+                logger.error("The message for the channel db should be a dictionary.")
         else:
-            print(f"The message was ignored because "
-                  f"the treatment for the channel {channelName} is not implemented.")
+            logger.error("The message was ignored because "
+                        "the treatment for the channel %s is not implemented.", channelName)
```
BECATRUE: Indentation!

examples/logger.py
```diff
@@ -186,10 +186,10 @@ def receivedSlot(self, channelName: str, content: Any):
             if isinstance(content, str):
                 self.addLog(content)
             else:
-                print("The message for the channel log should be a string.")
+                (logging.getLogger(__name__)).error("The message for the channel log should be a string.")
```
BECATRUE: Please define `logging.getLogger(__name__)` as a global variable like other files.

examples/logger.py
```diff
@@ -186,10 +186,10 @@ def receivedSlot(self, channelName: str, content: Any):
...
         else:
-            print(f"The message was ignored because "
-                  f"the treatment for the channel {channelName} is not implemented.")
+            (logging.getLogger(__name__)).error("The message was ignored because "
+                             "the treatment for the channel %s is not implemented.", channelName)
```
BECATRUE: Indentation!

examples/numgen.py
BECATRUE: Indentation!

examples/numgen.py
BECATRUE: Indentation!

examples/numgen.py
```diff
@@ -136,21 +141,19 @@ def receivedSlot(self, channelName: str, content: Any):
...
         self.dbName = self.generatorFrame.dbBox.currentText()
         self.viewerFrame.statusLabel.setText("database updated")
-        self.broadcast(
-            "log", 
-            f"Database to store is set as {self.dbName}." if self.dbName
-            else "Database to store is not selected."
-        )
+        if self.dbName:
+            logger.info("Database to store is set as %s", self.dbName)
+        else: logger.info("Database to store is not selected.")
```
BECATRUE: Please write the body statement to a new line.

examples/poller.py
```diff
@@ -122,27 +127,25 @@ def receivedSlot(self, channelName: str, content: Any):
...
         """Sets the database to store the polled number."""
         self.dbName = self.viewerFrame.dbBox.currentText()
-        self.broadcast(
-            "log", 
-            f"Polled database is set as {self.dbName}." if self.dbName
-            else "Polled database is not selected."
-        )
+        if self.dbName:
+            logger.info("Database to store is set as %s.", self.dbName)
+        else: logger.warning("Database to store is not selected.")
```
BECATRUE: New line!

Also, in `numgen.py`, you handled this case with `logger.info()`.
Please unify both statements of two files.

BECATRUE (changes requested):
Please check indentations!

examples/datacalc.py
Aijuh: sorry...

Aijuh:
I will re-read changed code for later. Sorry for repeated error.

BECATRUE:
> I will re-read changed code for later. Sorry for repeated error.

It's okay😃 I will review the updated parts.

examples/dbmgr.py
BECATRUE: How about using `logging.warning()`?
If you think `logging.info()` is more appropriate, never mind!

examples/poller.py
```diff
@@ -95,8 +100,8 @@ def updateDB(self, content: dict):
         newDBs = set([""])
         for db in content.get("db", ()):
             if any(key not in db for key in ("name", "path")):
-                print(f"The message was ignored because "
-                        f"the database {db} has no such key; name or path.")
+                logger.error("The message was ignored because "
+                            "the database %s has no such key; name or path.", json.dumps(db))
```
BECATRUE: Please check the indentation.

examples/poller.py
```diff
@@ -122,27 +127,26 @@ def receivedSlot(self, channelName: str, content: Any):
             if isinstance(content, dict):
                 self.updateDB(content)
             else:
-                print("The message for the channel db should be a dictionary.")
+                logger.error("The message for the channel db should be a dictionary.")
         else:
-            print(f"The message was ignored because "
-                  f"the treatment for the channel {channelName} is not implemented.")
+            logger.error("The message was ignored because "
+                        "the treatment for the channel %s is not implemented.", channelName)
```
BECATRUE: Here, too!

examples/logger.py
```diff
@@ -186,10 +189,10 @@ def receivedSlot(self, channelName: str, content: Any):
             if isinstance(content, str):
                 self.addLog(content)
             else:
-                print("The message for the channel log should be a string.")
+                logger.error("The message for the channel log should be a string.")
         else:
-            print(f"The message was ignored because "
-                  f"the treatment for the channel {channelName} is not implemented.")
+            logger.error("The message was ignored because "
```
BECATRUE: I just realized it now, and I think you can remove `receivedSlot()` becuase we don't use the `log` channel more.

BECATRUE (changes requested):
It looks almost done😃

I left only a few reviews.

examples/logger.py
Aijuh: I knew that but just leave it. I will delete receivedSlot.

BECATRUE: I'm sorry I should have told you earlier😅

Aijuh:
Code fixed.

examples/dbmgr.py
kangz12345: I think we can remove the header "DBMgrApp.openCloseDatacalc():" (for the other similar cases as well). Do you agree @BECATRUE?

examples/logger.py
Aijuh: Anytime.

examples/backend.py
```diff
@@ -56,7 +60,7 @@ def read(db_path: str, table: str) -> Any:
                 f"SELECT * FROM {table} ORDER BY rowid DESC LIMIT 1"
             ).fetchone()[0]
     except sqlite3.Error as e:
-        print(f"apps.backend.read(): {e!r}")
+        logging.exception("apps.backend.read(): %r", e)
```
kangz12345: Use `logger`, not `logging`.
Moreover, since `exception()` prints out the stack trace, you don't have to print the exception.
Instead, please describe what has failed.
For example: 
```python
logger.exception("Failed to read table %s from database %s.", table, db_path)
```

examples/dbmgr.py
```diff
@@ -121,10 +125,9 @@ def sendDB(self, isAdded: bool, name: str):
...
         msg = {"db": [db._asdict() for db in self.dbList]}
         self.broadcast("db", msg)
-        self.broadcast(
-            "log",
-            f"Database {name} is added." if isAdded
-            else f"Database {name} is removed."
+        logger.info(
+            "Database %s is added." if isAdded
+            else "Database %s is removed.", name
         )
```
kangz12345: Another way to write it:
```python
logger.info("Database %s is %s.", name, "added" if isAdded else "removed")
```
You can use your previous code! Just follow your preference.

examples/dbmgr.py
kangz12345: @BECATRUE @Aijuh

examples/logger.py
```diff
@@ -135,9 +138,9 @@ def __init__(self, name: str, parent: Optional[QObject] = None):
         fs ="%(name)s %(message)s"
         formatter = logging.Formatter(fs)
         self.handler.setFormatter(formatter)
-        logger = logging.getLogger()
-        logger.setLevel(logging.DEBUG)
-        logger.addHandler(self.handler)
+        rootLogger = logging.getLogger()
+        rootLogger.setLevel(logging.DEBUG)
+        rootLogger.addHandler(self.handler)
```
kangz12345: 👍

examples/logger.py
```diff
@@ -174,27 +177,10 @@ def addLog(self, content: str):
...
-                print("The message for the channel log should be a string.")
-        else:
-            print(f"The message was ignored because "
-                  f"the treatment for the channel {channelName} is not implemented.")
-
     @pyqtSlot()
     def checkToClear(self):
         """Shows a confirmation frame for log clearing."""
-        self.broadcast("log", "Clicked to clear logs")
+        logger.info("Clicked to clear logs")
```
kangz12345: Please append a period at the end.

examples/numgen.py
```diff
@@ -161,12 +165,13 @@ def generateNumber(self):
...
         # save the generated number
         dbPath = self.dbs[self.dbName]
         is_save_success = write(os.path.join(dbPath, self.dbName), self.table, num)
         if is_save_success:
             self.viewerFrame.statusLabel.setText("number saved successfully")
-            self.broadcast("log", "Generated number saved.")
+            logger.info("Generated number saved.")
         else:
             self.viewerFrame.statusLabel.setText("failed to save number")
+            logger.error("Failed to save generated number")
```
kangz12345: Please append a period at the end.

examples/poller.py
```diff
@@ -151,8 +155,10 @@ def poll(self):
...
         self.viewerFrame.numberLabel.setText(f"polled number: {num}")
-        self.broadcast("log", f"Polled number: {num}.")
+        logger.info("Polled number: %d.", num)
         # save the polled number
         dbPath = self.dbs[self.dbName]
         if write(os.path.join(dbPath, self.dbName), self.table, num):
-            self.broadcast("log", "Polled number saved.")
+            logger.info("Polled number saved.")
+        else:
+            logger.error("Failed to save polled number.")
```
kangz12345, line 164: 👍

kangz12345 (changes requested):

examples/dbmgr.py
Aijuh: I think this suggestion is easy to read. I will change in this way.

BECATRUE (changes requested):
Please apply the remaining parts that we discussed in snu-quiqcl/iquip#74.

kangz12345:
> Please apply the remaining parts that we discussed in [snu-quiqcl/iquip#74](https://github.com/snu-quiqcl/iquip/issues/74).

Please describe the request in more detail.

Aijuh:
I understand the request. I will update qiwis with respect to iquip.

Aijuh:
I will first update until currently merged `develop` branch of iquip. Give me comments if you want further update.

BECATRUE (dismissed):
LGTM!

Please request @kangz12345 for reviews.

examples/backend.py
```diff
@@ -99,8 +99,8 @@ def write(db_path: str, table: str, value: Any) -> bool:
                 f"INSERT INTO {table} VALUES (?, datetime('now', 'localtime'))", 
                 (value,)
             )
-    except sqlite3.Error as e:
-        logging.exception("apps.backend.write(): %r", e)
+    except sqlite3.Error:
+        logger.exception("Failed to write database %s from table %s.", db_path, table)
```
kangz12345: Since the table is in the database, this expression seems awkward.
How about `"Failed to write a value into table %s of database %s"`?

kangz12345 (changes requested):

examples/backend.py
Aijuh: I don't know details about database. I will fix it in this way.

Aijuh:
Fixed the code.

examples/backend.py
```diff
@@ -100,7 +100,7 @@ def write(db_path: str, table: str, value: Any) -> bool:
                 (value,)
             )
     except sqlite3.Error:
-        logger.exception("Failed to write database %s from table %s.", db_path, table)
+        logger.exception("Failed to write a value into table %s of database %s", db_path, table)
```
kangz12345: 😄 The order of the following arguments should be switched as well!

kangz12345 (changes requested):

Aijuh:
I fixed it. I should have read it.

kangz12345 (approved):
LGTM

BECATRUE (approved):
LGTM

PR #207, opened 2023-07-19
Handle re-creating the existing app

### LINKED ISSUE #190: [BUG] Handle re-creating the existing app
### Describe the bug
When an app _A_ exists, if we requests creating the app _A_ again, it seems to be added well on the GUI, but internally, the connection to the existing app is lost.

### Expected behavior
I came up with two solutions:
1. Ignore the incoming request
2. Replace the original app into the new one

### Additional context
For example, when you try to reopen a builder that is already open, how do you deal with it?

### LINKED ISSUE #199: Add log level selector

BECATRUE:
This PR will close #190.

I implement two things.
1. Add the param `replace` to handle the case where trying to re-create an existing app.
2. Update `createApp()` test and make `subscribe()` test

Suprisignly, we haven't tested for `subscribe()`, thus the coverage wasn't 100%.
I tried to implement it, but encountered a minor problem.

I should handle the case where the app tries to subscribe a channel already subscribed to.
But, I think there is no way to distinguish two cases (line 316 and 318), because we cannot mock `set.add()`.
Can I test two cases with mocking `logger`?

### DIFF
--- a/qiwis.py
+++ b/qiwis.py
@@ -232,7 +232,8 @@ class Qiwis(QObject):
         Args:
             name: The name of the app to be added.
             info: The AppInfo object describing the app.
-            replace: If True, the existing app will be replaced. Otherwise, nothing happens.
+            replace: It describes the behavior when trying to create an existing app.
+              If True, the original app will be replaced. Otherwise, it will be ignored.
         """
         if name in self._apps:
             if replace:
--- a/qiwis.py
+++ b/qiwis.py
@@ -226,13 +226,20 @@ class Qiwis(QObject):
         """Returns the names of the apps including whose frames are hidden."""
         return tuple(self._apps.keys())
 
-    def createApp(self, name: str, info: AppInfo):
+    def createApp(self, name: str, info: AppInfo, replace: bool = False):
         """Creates an app and shows their frames using set-up environment.
         
         Args:
-            name: A name of app.
-            info: An AppInfo object describing the app.
+            name: The name of the app to be added.
+            info: The AppInfo object describing the app.
+            replace: If True, the existing app will be replaced. Otherwise, nothing happens.
         """
+        if name in self._apps:
+            if replace:
+                self.destroyApp(name)
+            else:
+                logger.error("The app %s already exists.", name)
+                return
         with _add_to_path(os.path.dirname(info.path)):
             module = importlib.import_module(info.module)
         cls = getattr(module, info.cls)
--- a/test.py
+++ b/test.py
@@ -117,19 +117,13 @@ class QiwisTestWithApps(unittest.TestCase):
         app_.frames.return_value = (QWidget(),)
         cls = mock.MagicMock(return_value=app_)
         setattr(self.mocked_import_module.return_value, "cls2", cls)
+        appInfo = qiwis.AppInfo(module="module2", cls="cls2")
         # The original app will not be replaced.
-        self.qiwis.createApp(
-            "app2",
-            qiwis.AppInfo(module="module2", cls="cls2")
-        )
+        self.qiwis.createApp("app2", appInfo)
         mocked_destroy_app.assert_not_called()
         self.assertEqual(self.qiwis._apps["app2"], orgApp)
         # The original app will be replaced.
-        self.qiwis.createApp(
-            "app2",
-            qiwis.AppInfo(module="module2", cls="cls2"),
-            True
-        )
+        self.qiwis.createApp("app2", appInfo, True)
         mocked_destroy_app.assert_called_once_with("app2")
         self.assertNotEqual(self.qiwis._apps["app2"], orgApp)
 
--- a/test.py
+++ b/test.py
@@ -167,11 +167,15 @@ class QiwisTestWithApps(unittest.TestCase):
             )
 
     def test_subscribe(self):
+        self.assertNotIn("app1", self.qiwis._subscribers["ch3"])
         self.qiwis.subscribe("app1", "ch3")
         self.assertIn("app1", self.qiwis._subscribers["ch3"])
-        # The app already subscribes to the channel.
-        orgSubscribers = self.qiwis._subscribers["ch3"]
+    
+    def test_subscribe_duplicate(self):
+        """Tests for the case where trying to subscribe to a channel already subscribed to."""
         self.qiwis.subscribe("app1", "ch3")
+        orgSubscribers = self.qiwis._subscribers["ch3"]
+        self.qiwis.subscribe("app1", "ch3")  # Try to subscribe to the channel again.
         self.assertEqual(self.qiwis._subscribers["ch3"], orgSubscribers)
 
     def test_unsubcribe(self):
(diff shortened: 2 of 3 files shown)


qiwis.py
```diff
@@ -226,13 +226,21 @@ def appNames(self) -> Tuple[str]:
...
 
-    def createApp(self, name: str, info: AppInfo):
+    def createApp(self, name: str, info: AppInfo, replace: bool = False):
         """Creates an app and shows their frames using set-up environment.
         
         Args:
-            name: A name of app.
-            info: An AppInfo object describing the app.
+            name: The name of the app to be added.
+            info: The AppInfo object describing the app.
```
kangz12345, line 234: 👍

test.py
```diff
@@ -108,6 +108,31 @@ def test_create_app(self):
         self.assertIn("app3", self.qiwis._dockWidgets)
         self.assertIn("app3", self.qiwis._subscribers["ch1"])
 
+    @mock.patch("qiwis.Qiwis.destroyApp")
+    def test_create_existing_app(self, mockedDestroyApp):
```
kangz12345: We've used the snake case for the mocks: `mocked_destroy_app`.

test.py
```diff
@@ -108,6 +108,31 @@ def test_create_app(self):
...
+        orgApp = self.qiwis._apps["app2"]
+        app_ = mock.MagicMock()
+        app_.cls = "cls2"
+        app_.frames.return_value = (QWidget(),)
+        cls = mock.MagicMock(return_value=app_)
+        setattr(self.mocked_import_module.return_value, "cls2", cls)
+        # The original app will not be replaced.
+        self.qiwis.createApp(
+            "app2",
+            qiwis.AppInfo(**{"module": "module2", "cls": "cls2"})
```
kangz12345: Why don't you write: `qiwis.AppInfo(module="module2", cls="cls2")`?

test.py
```diff
@@ -108,6 +108,31 @@ def test_create_app(self):
...
+        )
+        mockedDestroyApp.assert_not_called()
+        self.assertEqual(self.qiwis._apps["app2"], orgApp)
+        # The original app will be replaced.
+        self.qiwis.createApp(
+            "app2",
+            qiwis.AppInfo(**{"module": "module2", "cls": "cls2"}),
```
kangz12345: How about assigning this to a variable since it is used twice.

kangz12345 (changes requested):
> But, I think there is no way to distinguish two cases (line 316 and 318), because we cannot mock set.add().

If you want to test both cases, you can mock the `self.qiwis._subscribers` dictionary.

> Can I test two cases with mocking logger?

I think it is not a good idea. Log messages or policy can be changed without any consideration of test codes.

BECATRUE:
> If you want to test both cases, you can mock the `self.qiwis._subscribers` dictionary.

It has a problem!
At the beginning of `subscribe()`, it checks if `app in self._subscribers[channel]` so that it shouldn't be mocked.
Another considerable solution is mocking `self.qiwis._subscribers[channel].add` function but it is blocked in Python.

test.py
BECATRUE: Oh, it looks clear!
I will apply this to other parts, too.

kangz12345:
> > If you want to test both cases, you can mock the `self.qiwis._subscribers` dictionary.
> 
> It has a problem! At the beginning of `subscribe()`, it checks if `app in self._subscribers[channel]` so that it shouldn't be mocked.

Oh, I was talking about `mock.patch.dict(self.qiwis._subscribers, {"ch": set()})` or something like that.

kangz12345:
> > > If you want to test both cases, you can mock the `self.qiwis._subscribers` dictionary.
> > 
> > 
> > It has a problem! At the beginning of `subscribe()`, it checks if `app in self._subscribers[channel]` so that it shouldn't be mocked.
> 
> Oh, I was talking about `mock.patch.dict(self.qiwis._subscribers, {"ch": set()})` or something like that.

I'm sorry now I get it. I will think about it.

kangz12345:
> > If you want to test both cases, you can mock the `self.qiwis._subscribers` dictionary.
> 
> It has a problem! At the beginning of `subscribe()`, it checks if `app in self._subscribers[channel]` so that it shouldn't be mocked. Another considerable solution is mocking `self.qiwis._subscribers[channel].add` function but it is blocked in Python.

How about exploiting the magic methods (note that `MagicMock` is `Mock` with magic methods implemented)?
For example, 
```python
m = mock.MagicMock()
m.__contains__.side_effect = ("a", "b").__contains__
```
Then `"a" in m` returns `True`, `"c" in m` returns `False`, and all the mocked calls are recorded as well.

BECATRUE:
Thanks for suggesting a solution😂

First, I solved the problem in my own way and please review it.

BECATRUE:
Please check the updated parts! @kangz12345

test.py
```diff
@@ -169,6 +169,10 @@ def test_subscriber_names(self):
     def test_subscribe(self):
         self.qiwis.subscribe("app1", "ch3")
         self.assertIn("app1", self.qiwis._subscribers["ch3"])
+        # The app already subscribes to the channel.
+        orgSubscribers = self.qiwis._subscribers["ch3"]
+        self.qiwis.subscribe("app1", "ch3")
+        self.assertEqual(self.qiwis._subscribers["ch3"], orgSubscribers)
```
kangz12345: Good! Why don't you separate this as another test method, e.g., `test_subscribe_duplicate()`?

test.py
```diff
@@ -169,6 +169,10 @@ def test_subscriber_names(self):
     def test_subscribe(self):
         self.qiwis.subscribe("app1", "ch3")
         self.assertIn("app1", self.qiwis._subscribers["ch3"])
```
kangz12345: How about asserting that `"app1"` was not in `_subscribers["ch3"]` before subscribing?

kangz12345 (changes requested):

test.py
BECATRUE: It looks more legible! I will do that.

test.py
BECATRUE: Good! I will apply it.

kangz12345 (approved):
LGTM

PR #211, opened 2023-08-03
Support a tab structure

### LINKED ISSUE #196: Support a tab structure
As I work on iquip project, I thought about supporing a tab structure in qiwis.
If we open several experiment builders, the GUI looks as below.

But I think showing the builders as a tab looks more legible using [`QTabWidgeet`](https://doc.qt.io/qtforpython-5/PySide2/QtWidgets/QTabWidget.html).

Feel free to leave your opinion😃

BECATRUE:
This PR will close #196.

I supported a tab structure, not showing in a row.

image

### DIFF
--- a/qiwis.py
+++ b/qiwis.py
@@ -203,7 +203,14 @@ class Qiwis(QObject):
             "bottom": Qt.BottomDockWidgetArea
         }.get(info.pos, Qt.LeftDockWidgetArea)
         if info.show:
-            self.mainWindow.addDockWidget(area, dockWidget)
+            areaDockWidgets = [
+                dockWidget_ for dockWidget_ in self.mainWindow.findChildren(QDockWidget)
+                if self.mainWindow.dockWidgetArea(dockWidget_) == area
+            ]
+            if areaDockWidgets:
+                self.mainWindow.tabifyDockWidget(areaDockWidgets[-1], dockWidget)
+            else:
+                self.mainWindow.addDockWidget(area, dockWidget)
         self._dockWidgets[name].append(dockWidget)
         logger.info("Added a frame to the app %s: %s", name, info)
 
--- a/test.py
+++ b/test.py
@@ -108,8 +108,7 @@ class QiwisTestWithApps(unittest.TestCase):
         self.assertIn("app3", self.qiwis._dockWidgets)
         self.assertIn("app3", self.qiwis._subscribers["ch1"])
 
-    @mock.patch("qiwis.Qiwis.destroyApp")
-    def test_create_existing_app(self, mocked_destroy_app):
+    def test_create_existing_app(self):
         """Tests for the case where trying to create an existing app."""
         orgApp = self.qiwis._apps["app2"]
         app_ = mock.MagicMock()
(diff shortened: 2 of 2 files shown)


BECATRUE:
I will modify the unittest after implementing the central widget.

qiwis.py
```diff
@@ -203,7 +203,14 @@ def addFrame(self, name: str, frame: QWidget, info: AppInfo):
             "bottom": Qt.BottomDockWidgetArea
         }.get(info.pos, Qt.LeftDockWidgetArea)
         if info.show:
-            self.mainWindow.addDockWidget(area, dockWidget)
+            areaDockWidgets = [
+                dockWidget_ for dockWidget_ in self.mainWindow.findChildren(QDockWidget)
```
kangz12345, line 207: 👍

qiwis.py
```diff
@@ -203,7 +203,14 @@ def addFrame(self, name: str, frame: QWidget, info: AppInfo):
             "bottom": Qt.BottomDockWidgetArea
         }.get(info.pos, Qt.LeftDockWidgetArea)
         if info.show:
-            self.mainWindow.addDockWidget(area, dockWidget)
+            areaDockWidgets = [
+                dockWidget_ for dockWidget_ in self.mainWindow.findChildren(QDockWidget)
+                if self.mainWindow.dockWidgetArea(dockWidget_) == area
+            ]
+            if areaDockWidgets:
+                self.mainWindow.tabifyDockWidget(areaDockWidgets[-1], dockWidget)
```
kangz12345, line 211: I think `areaDockWidgets[-1]` does not guarantee the order of tabbed widgets.
How about using [`tabifiedDockWidgets()`](https://doc.qt.io/qt-5/qmainwindow.html#tabifiedDockWidgets) to guarantee an intuitive order, e.g., the added tab is always comes to the front, etc.

kangz12345: If `areaDockWidgets` preserves the order of widget addition, it would be okay.

BECATRUE: > If areaDockWidgets preserves the order of widget addition, it would be okay.

Umm.. I think it doesn't guarantee the order😂

BECATRUE: Now, I realize that `tabifyDockWidget()` always adds the `second` widget on the top of the area of the `first` widget.
(Here, the top means the last widget.)

For test, I changed it into `areaDockWidgets[0]`, but nothing was changed.

kangz12345: Oh, good observation! Then never mind.

BECATRUE:
Then, is there something to modify? @kangz12345

Aijuh (dismissed):
I understand it as giving tab structure in the mainWindow. Not showing all the widget in one frame. 
Even if it has unittest error, I think you can talk to @kangz12345 about it. I will first approve it.

kangz12345:
> I will modify the unittest after implementing the central widget.

Is the unit test error expected?

BECATRUE:
> Is the unit test error expected?

No😮 I realized it just before and tried to fix it, but only unittest for Python 3.11 was failed.

BECATRUE:
I found out the soluting using `patch.object()`.

kangz12345:
Why was it a problem only in 3.11? Do you know the exact cause?

BECATRUE:
> Why was it a problem only in 3.11? Do you know the exact cause?

Oh, I resolved it! I don't know the exact reason, but it looks like I did the wrong mocking.

BECATRUE:
Please check this! @kangz12345 @Aijuh

kangz12345:
I have 2 questions:
1. When we create an app at e.g. bottom and there were several frames (several columns), which one is tabified?
2. When we destroy an app whose frame was at e.g. bottom and there were 2 frames, will the remaining 1 frame be in a tab or not?

BECATRUE:
> When we create an app at e.g. bottom and there were several frames (several columns), which one is tabified?

In my simple test, a new frame is created in the frame with the latest added frame, because I used `tabifyDockWidget()` with the last element of `areaDockWidgets` as follows:
```python
self.mainWindow.tabifyDockWidget(areaDockWidgets[-1], dockWidget)
```

If I use the first element `areaDockWidgets[0]`, the frame is created in the frame with the first added frame as we expect.

BECATRUE:
> When we destroy an app whose frame was at e.g. bottom and there were 2 frames, will the remaining 1 frame be in a tab or not?

The remaining one is not in a tab, but just a single docked widget.
It seems that PyQt doesn't support a tab with only a frame.

kangz12345:
Thank you for the confirmation!

kangz12345 (approved):
LGTM

BECATRUE:
Please re-check this PR! @Aijuh

Aijuh (approved):

PR #213, opened 2023-08-19
Support a central widget

### LINKED ISSUE #205: Support a central widget
Currently, we can add an app into only four positions; top, left, right, bottom.

However, I think we need a central widget, such as a builder in the artiq-dashboard.

image

Thus, the central position will be supported in this issue.

BECATRUE:
This PR will close #205.

Now, a central widget is supported. The impletations are as follows:
1. The central area is a `QMdiArea` instance.
2. In this area, a `QMdiSubWindow` can be added.
3. If the `pos` argument is set to "center" in `config.json`, each widget of the app will be wrapped in `QMdiSubWindow`.

For test, I modified the position of `datacalc` app to "center".

image

### DIFF
--- a/qiwis.py
+++ b/qiwis.py
@@ -60,18 +60,21 @@ class AppInfo(Serializable):
     """Information required to create an app.
     
     Fields:
-        module: Module name in which the app class resides.
-        cls: App class name.
-        path: System path for module importing.
-        show: Whether to show the app frames on creation.
-        pos: Position on the main window; refer to Qt.DockWidgetArea enum.
-          Should be one of "left", "right", "top", or "bottom", case-sensitive.
-          Otherwise, defaults to Qt.LeftDockWidgetArea.
-        channel: Channels which the app subscribes to.
-        args: Keyword argument dictionary of the app class constructor.
-          It must exclude name and parent arguments. Even if they exist, they will be ignored.
+        module: The module name of the app class.
+        cls: The name of the app class.
+        path: The path for importing the module.
+        show: Whether to show frames of the app.
+        pos: The position of the frames on the GUI.
+          It should be one of "center", "left", "right", "top", or "bottom", case-sensitive.
+          Otherwise, it is set to "left".
+          In the case "center", the frame is wrapped by QMdiSubWindow.
+          In the other cases, the frame is wrapped by QDockWidget and
+            its position follows Qt.DockWidgetArea.
+        channel: The list of channels which the app subscribes to.
+        args: The dictionary for the keyword arguments of the app class constructor.
+          It should exclude the name and parent arguments.
           None for initializing the app with default values,
-          where only name and parent arguments will be passed.
+            where only the name and parent arguments will be passed.
     """
     module: str
     cls: str
--- a/qiwis.py
+++ b/qiwis.py
@@ -202,6 +202,7 @@ class Qiwis(QObject):
             outerWidget = QMdiSubWindow(self.centralWidget)
             outerWidget.setWindowTitle(name)
             outerWidget.setWidget(frame)
+            outerWidget.show()
         else:
             outerWidget = QDockWidget(name, self.mainWindow)
             outerWidget.setWidget(frame)
--- a/qiwis.py
+++ b/qiwis.py
@@ -199,13 +199,13 @@ class Qiwis(QObject):
             info: An AppInfo object describing the app.
         """
         if info.pos == "center":
-            outerWidget = QMdiSubWindow(self.centralWidget)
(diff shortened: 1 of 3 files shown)


qiwis.py
```diff
@@ -161,10 +166,9 @@ def __init__(
         super().__init__(parent=parent)
         self.appInfos = appInfos
         self.mainWindow = QMainWindow()
-        self.centralWidget = QWidget()
-        self.centralWidget.setStyleSheet("border-image: url(./resources/background.png);")
+        self.centralWidget = QMdiArea()
         self.mainWindow.setCentralWidget(self.centralWidget)
-        self._dockWidgets = defaultdict(list)
+        self._outerWidgets = defaultdict(list)
```
kangz12345: What is the meaning of "outer widget"? I don't get the intuition from its name.
If there is no special reason, how about use the term "frame"?
For example, `self._frames`.
Also you can define a new type annotation for convenience: `Frame = Union[QMdiSubWindow, QDockWidget]`.
However, it is true that "frame" is not so precisely defined.

BECATRUE: I mean, it's a role to wrap up something.
Also, I agree with you that "frame" is not exact because there is a frame in `QMdiSubWindow` or `QDockWidget`, not in itself.

Thus, how about `self._wrapperWidgets`? I want to use "widget" because the common parent class of `QMdiSubWindow` and `QDockWidget` is `QWidget`.

kangz12345: Okay then, please rename it to wrapper widgets!

BECATRUE:
I updated it. @kangz12345

qiwis.py
```diff
@@ -194,39 +198,48 @@ def addFrame(self, name: str, frame: QWidget, info: AppInfo):
...
-                dockWidget_ for dockWidget_ in self.mainWindow.findChildren(QDockWidget)
-                if self.mainWindow.dockWidgetArea(dockWidget_) == area
-            ]
-            if areaDockWidgets:
-                self.mainWindow.tabifyDockWidget(areaDockWidgets[-1], dockWidget)
-            else:
-                self.mainWindow.addDockWidget(area, dockWidget)
-        self._dockWidgets[name].append(dockWidget)
+        if info.pos == "center":
+            outerWidget = QMdiSubWindow(self.centralWidget)
```
kangz12345: I think this (and any other `outerWidget`) should be changed to `wrapperWidget` as well.

test.py
```diff
@@ -131,26 +131,26 @@ def test_destroy_app(self):
...
             self.assertNotIn(name, self.qiwis._apps)
-            self.assertNotIn(name, self.qiwis._dockWidgets)
+            self.assertNotIn(name, self.qiwis._wrapperWidgets)
             for channel in info.channel:
                 self.assertNotIn(name, self.qiwis._subscribers[channel])
 
     def test_update_frames_inclusive(self):
         """Tests for the case where a new frame is added in the return of frames()."""
-        orgFramesSet = {dockWidget.widget() for dockWidget in self.qiwis._dockWidgets["app1"]}
+        orgFramesSet = {dockWidget.widget() for dockWidget in self.qiwis._wrapperWidgets["app1"]}
```
kangz12345: `wrapperWidget` instead of `dockWidget` 😄

kangz12345 (changes requested):

test.py
BECATRUE: I'm sorry I forgot a lot of things😥
I will check it again.

qiwis.py
```diff
@@ -194,39 +198,48 @@ def addFrame(self, name: str, frame: QWidget, info: AppInfo):
...
+            area = {
+                "left": Qt.LeftDockWidgetArea,
+                "right": Qt.RightDockWidgetArea,
+                "top": Qt.TopDockWidgetArea,
+                "bottom": Qt.BottomDockWidgetArea
+            }.get(info.pos, Qt.LeftDockWidgetArea)
+            if info.show:
+                areaDockWidgets = [
+                    dockWidget_ for dockWidget_ in self.mainWindow.findChildren(QDockWidget)
+                    if self.mainWindow.dockWidgetArea(dockWidget_) == area
```
kangz12345: Oops, I'm sorry for the late review. Now there is not `dockWidget`, so you don't have to put the trailing underscore `dockWidget_`.

kangz12345 (changes requested):
Final!

kangz12345 (approved):
LGTM

BECATRUE:
Please check this! @Aijuh

Aijuh (approved):
LGTM

PR #227, opened 2023-09-27
Remove show option & Support floating area

### LINKED ISSUE #210: Handle a frame with the initial `False` show option
So far, we have supported the `show` option in `config.json`.

Only if the option is `True`, frames of the app are shown.
However, in the current system, we cannot show frames with the initial `False` show option, thus, these frames have become obsolete.

Thus, we should determine how to handle this option.

### LINKED ISSUE #226: New initial position "floating"
Add a new initial frame position "floating", which is not docked but floating when initialized.

First create a dock widget and add it to the main window (docked). Then call `setFloating(True)`.

BECATRUE:
This closes #210 and closes #226.

These two issues are simple, so I implemented them together.
If you think it should be separated, I will do so.

### Screenshot
image

Of course, a docked widget can be floating, and the opposite is possible.

### DIFF
--- a/examples/config.json
+++ b/examples/config.json
@@ -3,7 +3,6 @@
         "numgen": {
             "module": "examples.numgen",
             "cls": "NumGenApp",
-            "show": true,
             "pos": "left",
             "channel": ["db"],
             "args": {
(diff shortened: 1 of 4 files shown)


kangz12345 (approved):
LGTM!

PR #233, opened 2023-10-12
Destroy "center" apps when closed

### LINKED ISSUE #232: Destroy MDI area apps when closed
See snu-quiqcl/iquip#137 for details.

kangz12345:
This closes #232.

This implementation is simple but does not cover all cases.
For example, when an app has multiple frames, then the app will be destroyed even if just one frame is closed.
However, the current frame-management system is not that flexible and we should do some work on it.
I think it is okay for now to keep this implementation, but please let me know if you think different!

### DIFF
--- a/qiwis.py
+++ b/qiwis.py
@@ -31,7 +31,7 @@ from typing import (
 )
 
 from PyQt5.QtCore import QObject, pyqtSignal, pyqtSlot, Qt
-from PyQt5.QtGui import QIcon, QPainter, QPaintEvent, QPixmap
+from PyQt5.QtGui import QIcon, QPainter, QPaintEvent, QPixmap, QCloseEvent
 from PyQt5.QtWidgets import (
     QApplication, QDockWidget, QMainWindow, QMdiArea, QMdiSubWindow, QMessageBox, QWidget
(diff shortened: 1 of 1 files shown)


BECATRUE:
I also agree with you, that the current frame management system has several problems.

But, I think removing only clicked frame may cause critical problems.
In my idea, in order to remove all frames of the clicked app, it is okay to follow these steps:
1. Connect `closed` signal to `destroyApp()`
2. In [removeFrame()](https://github.com/snu-quiqcl/qiwis/pull/233/files#diff-27a47b106eb8b3db7f941ca6297a2273c3c46f309b4f79b900fb1e7d58a40f29R300), call `frame.closeEvent()`.

How about this way?

kangz12345:
> But, I think removing only clicked frame may cause critical problems

Could you explain more about this?
Currently I connected `destroyApp()` to `closed` signal, so all the frames will be removed.

BECATRUE:
Oh, I couldn't check the signal connection code.. Sorry for confusing. There is no problem!

BECATRUE (approved):
If you tested if it works well, LGTM!

PR #238, opened 2023-10-25
Support for frame title

### LINKED ISSUE #231: Frame titles
For multi-frame apps, the frames have the same title, so it is hard to distinguish them when they are in tabs.
I think we should support individual frame title naming.

TBC

### Additional context
In `iquip`, the device monitor app has three frames and they all have the same title "monitor".

BECATRUE:
This closes #231.

Now, a frame title is supported and it can be set in `frames()` of each app.

To meet the modified `frames()`, I updated `Qiwis` class and several unit tests.

### DIFF
--- a/qiwis.py
+++ b/qiwis.py
@@ -569,11 +569,12 @@ class BaseApp(QObject):
         """The global constant namespace."""
         return self._constants
 
-    def frames(self) -> Iterable[QWidget]:
+    def frames(self) -> Iterable[Tuple[str, QWidget]]:
         """Gets frames for which are managed by the App.
 
         Returns:
-            An iterable object of Frame objects for showing.
+            An iterable object with frames info for showing.
+            Each entry is a tuple with frame title and frame object.
         """
         return ()
 
--- a/qiwis.py
+++ b/qiwis.py
@@ -378,7 +378,8 @@ class Qiwis(QObject):
         orgFramesSet = set(orgFrames)
         newFramesSet = set(newFrames)
         for frame in orgFramesSet - newFramesSet:
-            self.removeFrame(name, orgFrames[frame])
+            wrapperWidget = orgFrames[frame]
+            self.removeFrame(name, wrapperWidget)
         for frame in newFramesSet - orgFramesSet:
             title = newFrames[frame]
             self.addFrame(name, title, frame, info)
--- a/qiwis.py
+++ b/qiwis.py
@@ -570,7 +570,7 @@ class BaseApp(QObject):
         return self._constants
 
     def frames(self) -> Iterable[Tuple[str, QWidget]]:
-        """Gets frames for which are managed by the App.
+        """Returns frames info for showing.
 
         Returns:
             An iterable object with frames info for showing.
--- a/qiwis.py
+++ b/qiwis.py
@@ -370,18 +370,18 @@ class Qiwis(QObject):
         """
         app = self._apps[name]
         info = self.appInfos[name]
-        orgFrames = {
+        wrapperWidgets = {
             wrapperWidget.widget(): wrapperWidget
             for wrapperWidget in self._wrapperWidgets[name]
         }
-        newFrames = {frame: title for title, frame in app.frames()}
-        orgFramesSet = set(orgFrames)
-        newFramesSet = set(newFrames)
+        frameTitles = {frame: title for title, frame in app.frames()}
+        orgFramesSet = set(wrapperWidgets)
+        newFramesSet = set(frameTitles)
         for frame in orgFramesSet - newFramesSet:
-            wrapperWidget = orgFrames[frame]
+            wrapperWidget = wrapperWidgets[frame]
             self.removeFrame(name, wrapperWidget)
         for frame in newFramesSet - orgFramesSet:
-            title = newFrames[frame]
+            title = frameTitles[frame]
             self.addFrame(name, title, frame, info)
         logger.info("Updated frames: %d -> %d", len(orgFramesSet), len(newFramesSet))
 
--- a/qiwis.py
+++ b/qiwis.py
@@ -255,7 +255,7 @@ class Qiwis(QObject):
         
         Args:
             name: The app name to which frame is added.
-            titme: The frame title.
+            title: The frame title.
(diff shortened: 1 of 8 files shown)


qiwis.py
```diff
@@ -248,24 +248,25 @@ def load(self, appInfos: Mapping[str, AppInfo]):
...
         """Adds the given frame and wraps it with a wrapper widget.
 
         This is not a qiwiscall because QWidget is not Serializable.
         
         Args:
-            name: A name of app.
-            frame: A frame to show.
-            info: An AppInfo object describing the app.
+            name: The app name to which frame is added.
+            titme: The frame title.
```
kangz12345: Typo: `titme`.

qiwis.py
```diff
@@ -248,24 +248,25 @@ def load(self, appInfos: Mapping[str, AppInfo]):
...
+            frame: Frame object to show.
```
kangz12345: Since we didn't define `Frame` as a class, this expression looks awkward a little.
How about e.g. `Frame widget to show.`?

qiwis.py
```diff
@@ -248,24 +248,25 @@ def load(self, appInfos: Mapping[str, AppInfo]):
...
+            info: AppInfo object describing the app.
         """
         if info.pos == "center":
             wrapperWidget = MdiSubWindow(self.centralWidget)
-            wrapperWidget.setWindowTitle(name)
+            wrapperWidget.setWindowTitle(title)
```
kangz12345: How about keeping the app name?
For example, `f"{name} - {title}"`, `f"{title} ({name})"`, etc.
I think we should be able to know the app name of each frame.

kangz12345: In addition, it would be good to have an empty title - whose actual title is just the app name.

qiwis.py
```diff
@@ -362,24 +363,26 @@ def destroyApp(self, name: str):
...
+            name: The app name to update its frames.
         """
         app = self._apps[name]
         info = self.appInfos[name]
         orgFrames = {
             wrapperWidget.widget(): wrapperWidget
             for wrapperWidget in self._wrapperWidgets[name]
         }
-        newFrames = app.frames()
+        newFrames = {frame: title for title, frame in app.frames()}
```
kangz12345: Now both are dictionary, so it looks quite confusing.
How about `wrapperWidgets` and `frameTitles`, leaving the sets?

test.py
```diff
@@ -135,20 +135,24 @@ def test_destroy_app(self):
...
-        orgFramesSet = {wrapper.widget() for wrapper in self.qiwis._wrapperWidgets["app1"]}
+        orgFramesSet = {wrapperWidget.widget()
+                        for wrapperWidget in self.qiwis._wrapperWidgets["app1"]}
         newFramesSet = orgFramesSet | {QWidget()}
-        self.qiwis._apps["app1"].frames.return_value = tuple(newFramesSet)
+        self.qiwis._apps["app1"].frames.return_value = (("title", frame) for frame in newFramesSet)
         self.qiwis.updateFrames("app1")
-        finalFramesSet = {wrapper.widget() for wrapper in self.qiwis._wrapperWidgets["app1"]}
+        finalFramesSet = {wrapperWidget.widget()
+                          for wrapperWidget in self.qiwis._wrapperWidgets["app1"]}
```
kangz12345: I think `wrapper` is enough, but you can change it if you want to.

test.py
```diff
@@ -135,20 +135,24 @@ def test_destroy_app(self):
 
     def test_update_frames_inclusive(self):
         """Tests for the case where a new frame is added in the return of frames()."""
-        orgFramesSet = {wrapper.widget() for wrapper in self.qiwis._wrapperWidgets["app1"]}
+        orgFramesSet = {wrapperWidget.widget()
+                        for wrapperWidget in self.qiwis._wrapperWidgets["app1"]}
         newFramesSet = orgFramesSet | {QWidget()}
-        self.qiwis._apps["app1"].frames.return_value = tuple(newFramesSet)
+        self.qiwis._apps["app1"].frames.return_value = (("title", frame) for frame in newFramesSet)
```
kangz12345: It is often a better practice to _fix_ the generator values unless it requires large memory.
For example if `newFramesSet` is modified after defining this generator, and then we consume its values, the values are generated based on the modified `newFramesSet`.
Since this is a test code I think using `tuple(...)` to fix the values would be better!

kangz12345: An interesting demonstration: https://stackoverflow.com/questions/70342301/closure-semantics-in-python-generator-comprehensions

kangz12345 (changes requested):

qiwis.py
BECATRUE: Oh, it looks more reasonable!

qiwis.py
BECATRUE: I see! Then, I will adopt the first suggestion, but if an empty title, I will show only the app name.

test.py
BECATRUE: I also considered to use `wrapper`, but as it is a Python grammatical term, so I remained it.
However, you said okay, I will change it!

qiwis.py
BECATRUE: Could you explain more? I didn't understand it..

kangz12345: For those `dict`s, how about renaming `orgFrames` and `newFrames` e.g. `wrapperWidgets` and `frameTitles`, respectively?

test.py
BECATRUE: Wow, your example is too surprising! Then, I will use `tuple(...)` directly.

BECATRUE: Umm.. is it sufficient that `(tuple("title", frame) for frame in newFramesSet)`?

qiwis.py
BECATRUE: Oh, I see! It looks more legible, so I will adopt it.

test.py
kangz12345: You should write `tuple` outside: `tuple(("title", frame) for ...)`.

BECATRUE: I see! 👍

kangz12345 (approved):
LGTM

PR #240, opened 2023-11-17
Add `trust` opt in app info to bypass checking permission

### LINKED ISSUE #237: Bothersome by the message box in qiwis-call
When handling a qiwis-call, it always requests confirmation by the message box, hence it is too bothering.

How about removing this procedure or checking only specified apps in config file?

BECATRUE:
This closes #237.

I set `trust` option of `numgen` app to `True`, so the qiwiscall requested by it are bypassed to check permission.

### DIFF
--- a/examples/config.json
+++ b/examples/config.json
@@ -5,6 +5,7 @@
             "cls": "NumGenApp",
             "pos": "left",
             "channel": ["db"],
+            "trust": true,
             "args": {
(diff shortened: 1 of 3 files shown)


BECATRUE:
Now, all actions work well! @kangz12345

kangz12345 (approved):
LGTM

PR #244, opened 2023-11-22
Fix the bug that shows frame titles incorrectly

### LINKED ISSUE #243: Make frame titles shown correctly
### Describe the bug
When implementing frame titles in #231, I made a mistake that updates only center frames, except for left, right, top, bottom frames.
Thus, if we set a frame title to an empty string, we intended to show `{app name}`, but it becomes just an empty string.

To fix it, it looks okay to modify this line in [qiwis.py](https://github.com/snu-quiqcl/qiwis/blob/develop/qiwis.py#L271).

BECATRUE:
This closes #243.

To release the modifed version quickly, I updated version name in README and `setup.py`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ QIWIS (**Q**u**I**qcl **W**idget **I**ntegration **S**oftware) is a framework fo
 ## How to download and install
 One of the below:
 - Using released version (_recommended_)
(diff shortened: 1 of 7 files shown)


kangz12345 (approved):
LGTM!

PR #250, opened 2024-04-07
Improve background of `QMdiArea`

### LINKED ISSUE #249: Improve background of `QMdiArea`
So far we only support the background image of `QMdiArea`.
However, it has some conveniences in setting the background.
So I will implement to set the background color.

The background color will be given in the config file.

BECATRUE:
This closes #249.

What I have implemented is:
- Support the background color.
- Upgrade the mechanism drawing the background.

image


qiwis.py
```diff
@@ -145,28 +145,32 @@ class MdiArea(QMdiArea):
...
         """
-        QMdiArea.paintEvent(self, event)
         painter = QPainter(self.viewport())
-        x = (self.width() - self.background.width()) // 2
-        y = (self.height() - self.background.height()) // 2
-        painter.drawPixmap(x, y, self.background)
+        if self.background_color is not None:
+            painter.fillRect(self.rect(), QColor(int(self.background_color, 16)))
+        if self.background_path is not None:
+            image = QPixmap(self.background_path)
```
kangz12345: Is this efficient enough? It looks like it loads the image for every paint event.

BECATRUE: Oh, you're right! This method is called whenever the size of `QMdiArea` is changed.
I will improve it not to load the image every time.

qiwis.py
```diff
@@ -213,36 +216,29 @@ def __init__(
         """
         super().__init__(parent=parent)
         self.appInfos: Dict[str, AppInfo] = {}
-        icon_path, background_path = map(
-            lambda attr: getattr(constants, attr, None),
-            ("icon_path", "background_path")
+        icon_path, background_path, background_color = map(
+            lambda name, default: getattr(constants, name, default),
+            (("icon_path", ""), ("background_path", ""), ("background_color", "ffffff"))
```
kangz12345: Each tuple, e.g., ("icon_path", ""), is given as `name`, which raises the error that `default` is not given.
How about just using comprehension?
```python
... = (
  getattr(constants, name, default) for name, default in
  (("icon_path", ""), ("background_path", ""), ("backgroun_color", "ffffff"))
)
```

kangz12345 (changes requested):

qiwis.py
BECATRUE: Oh, I also found this problem and was thinking of a solution.
I think your solution is the most compact! I will apply it.

kangz12345 (approved):
LGTM

PR #252, opened 2024-04-08
Use `requirements.txt` instead of direct installation

### LINKED ISSUE #251: Use `requirements.txt` instead of direct installation
To manage easily, we need to use a requirements file instead of direct installation in GitHub actions.

BECATRUE:
This closes #251.

After merged, I will release a new version, v3.0.2.


BECATRUE:
Which is better indicating each specific version or only library names?

The first helps to re-create the same environment, but even several submodules are shown, making it less legible.
The second is legible, but it doesn't guarantee the same environment so later it may not work normally, like our PID GUI.

kangz12345:
> Which is better indicating each specific version or only library names?

You mean the requirements.txt?
I prefer the former.
However, is it possible to omit the minor versions, e.g., `1.2.x`?
Too strict version requirement might cause dependency issues when many requirements come together.

BECATRUE:
> You mean the requirements.txt?

Yes!

> is it possible to omit the minor versions, e.g., 1.2.x?

Sure. You can find the format of `requirements.txt` in [here](https://pip.pypa.io/en/stable/reference/requirements-file-format/).
Also, as you suggested, I will indicate the major(1st) and minor(2nd) version, not the patch(3rd) version.

BECATRUE:
I updated it just now. @kangz12345

kangz12345 (approved):
Cool!

